PR #5, opened 2018-11-27
fix: startup database error

michael1011:
This PR fixes an startup error I introduces in #2 when enabling `declaration` in `tsconfig.json`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "boltz-backend",
-  "version": "1.0.0",
+  "version": "0.0.1",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "boltz-backend",
-  "version": "1.0.0",
+  "version": "0.0.1",
```
ImmanuelSegol: Did we want this `1.0.0` or  `0.0.1`?

michael1011: We want `0.0.1`. It looks like i forgot to run `npm i` when changing it. But this shouldn't matter too much since it is just the lock file

PR #29, opened 2018-12-15
refactor(Config.ts): parse currencies

ImmanuelSegol:
The cli input must be in the following format due to how the cli parses the strings:
`'{ "symbol": "BTC", "network": "bitcoinTestnet", "chain": { "host": "127.0.0.1", "port": 18334, "certpath": "service_data_dir", "rpcuser": "user", "rpcpass": "user" }, "lnd": { "host": "128.0.0.1", "port": 10009, "certpath":"tls.cert", "macaroonpath":"lnd_data_chain_admin.macaroon" } }'`

I was thinking of adding the ability to just state the services you wish to use and we will get the settings based on the data directory. This would help in a situation where a user has `btcd`, `bitcoin-core`, and `parity-bitcoin` installed and giving them the option of choosing easily.


michael1011:
> The cli input must be in the following format due to how the cli parses the strings:

I think that the currencies array should be given to Boltz via a config file instead of a CLI argument. At least with the current approach it makes no sense to have such a long string as an argument.

> I was thinking of adding the ability to just state the services you wish to use and we will get the settings based on the data directory. This would help in a situation where a user has btcd, bitcoin-core, and parity-bitcoin installed and giving them the option of choosing easily.

Could you elaborate a bit? What do you mean exactly? Also keep in mind that some people are running their nodes with command line arguments that can't be parsed from the data directory.

bin/boltzd
```diff
@@ -43,6 +48,22 @@ const { argv } = require('yargs')
       describe: 'Path to the key of the TLS certificate of the Boltz gRPC interface',
       type: 'string',
     },
+    'btcd.path': {
+      describe: 'Path to btcd config',
+      type: 'string',
+    },
+    'btcd.lndlisten': {
```
michael1011: I would call the lnd options to avoid confusion:

- `lndbtc`
- `lndltc`

Also I think you meant the host and port for the gRPC interface instead of the `Uri` which is used for connecting to the node. And why is there no equivalent of `path` for the LNDs?

bin/boltzd
```diff
@@ -43,6 +48,22 @@ const { argv } = require('yargs')
       describe: 'Path to the key of the TLS certificate of the Boltz gRPC interface',
       type: 'string',
     },
+    'btcd.path': {
```
michael1011: What about calling all paths to a config file `configpath` so that it is consistent?

lib/Utils.ts
```diff
@@ -213,6 +213,15 @@ export const getScriptHashEncodeFunction = (outputType: OutputType) => {
   }
 };
 
+export const symbolToChain = (symbol: Symbols): Chain => {
```
michael1011: You could use the `Symbols` enum as identifier of the `Chain` enum and avoid having to use this method

lib/consts/Enums.ts
```diff
@@ -8,6 +8,23 @@ export enum ErrorCodePrefix {
   Swap = 6,
 }
 
+export enum Symbols {
```
michael1011: The name of enums should always be singular 

```suggestion
export enum Symbol {
```

bin/boltzd
```diff
@@ -43,6 +48,26 @@ const { argv } = require('yargs')
       describe: 'Path to the key of the TLS certificate of the Boltz gRPC interface',
       type: 'string',
     },
+    'btcd.configpath': {
+      describe: 'Path to btcd config',
+      type: 'string',
+    },
+    'btcd.lnd': {
```
michael1011: This is the host and port of the LND that is connected to BTCD? Right? If so we should first of all be consistent with the rest of the config and have two separate fields for port and host and call it `lndbtc.host` and `lndbtc.port`. What do you think?

lib/Config.ts
```diff
@@ -57,6 +58,7 @@ class Config {
 
       datadir: this.dataDir,
       loglevel: this.getDefaultLogLevel(),
+      network: Network.Testnet,
```
michael1011: I think `simnet` should be default since that one is used most of the time in development. Maybe we could do something similar to `getDefaultLogLevel`: if `process.env.NODE_ENV === 'production'` set it to testnet by default. if not to simnet?

bin/boltzd
```diff
@@ -43,6 +48,26 @@ const { argv } = require('yargs')
...
+    },
+    'ltcd.configpath': {
+      describe: 'Path to ltcd config',
+      type: 'string',
+    },
+    'ltcd.lnd': {
+      describe: 'Host and port for the gRPC interface',
+      type: 'string',
+    },
+    lndpath: {
```
michael1011: I am using two different data directories for the BTC and LTC LND. But I think people should use the config for non default configurations?

ImmanuelSegol: @michael1011 Yeah definitely if a user plans on having a complex setup he should use the conf file.
Especially later on when we will support multiple chains, it wouldn't be realistic to config it all through the CLI. Do you agree?

michael1011: Yeah completely.

PR #57, opened 2019-01-08
feat: add timeout height to grpc swap commands

### LINKED ISSUE #42: Make timeout of Swap scripts configurable
The timeout for the Swap scripts is `bestBlockHeight + 10` right now. To make that configurable the RPC methods `CreateSwap` and `CreateReverseSwap` should take some kind of timeout as parameter.

I would suggest a relative block height instead of an absolute. So that if the middleware sets `timeoutBlockHeight = x` the backend gets the height of the best block and adds `x` to it.

Also the default timeout should be configurable for each chain. 10 blocks on Bitcoin take a longer amount of time than 10 blocks on Litecoin or Ethereum statistically speaking.

ImmanuelSegol:
Closes #42


lib/cli/commands/CreateSwap.ts
```diff
@@ -21,6 +22,10 @@ export const builder = {
     describe: 'public key with which a refund transaction has to be signed',
     type: 'string',
   },
+  timeout_block_height: {
```
michael1011: This property exists twice and therefore should be moved to `BuilderComponents`

lib/cli/commands/CreateReverseSwap.ts
```diff
@@ -21,6 +21,10 @@ export const builder = {
     describe: 'amount of the invoice that will be returned',
     type: 'number',
   },
+  timeout_block_height: {
```
michael1011: I meant that you should rename `timeout_block_height` to `timeout_block_number` everywhere (also in the proto file and forget about the comment there).

lib/cli/commands/CreateSwap.ts
```diff
@@ -21,6 +22,11 @@ export const builder = {
     describe: 'public key with which a refund transaction has to be signed',
     type: 'string',
   },
+  timeout_block_number: {
+    describe: 'block height timeout',
+    type: 'number',
+    default: 10,
```
michael1011: Also the default should be in the Service class so that it affects all calls and not only the CLI ones

michael1011: The default value should be in the `SwapManager` instead of this CLI command

PR #58, opened 2019-01-08
feat: estimate fee using the average fee in the mempool

### LINKED ISSUE #36: Find solution for fee estimation
LTCD does not support the [`estimateFee`](https://godoc.org/github.com/btcsuite/btcd/rpcclient#Client.EstimateFee) method of BTCD. I discovered this in #35 and just hardcoded a fee of 10 sat per vbyte to avoid `not implemented` errors by LTCD.

Now we have to find a way to estimate a fee to use. I can think of two possible solutions:
- use another full node implementation that has fee estimation methods for BTC and LTC. And since that full node implementation has to be compatible with LND the only option is Bitcoin Core and Litecoin Core
- implement our own fee estimation tool by iterating over the transaction in the latest blocks and in the mempool to determine a reasonable fee

(Or we could simply not care about fees in Litecoin since the blocks are never full but that is not a long term solution.)

What do you think? @ImmanuelSegol @peartobear

ImmanuelSegol:
Closes #36


lib/chain/ChainClient.ts
```diff
@@ -105,6 +109,23 @@ class ChainClient extends BaseClient implements ChainClientInterface {
...
+  public estimateFee = () => {
+    return new Promise(async (res, rej) => {
+      try {
+        let sumOfFees = 0;
+        const mempoolTx = await this.getRawMempool(true);
+        const transactions = Object.values(mempoolTx);
+        transactions.forEach((tx: any) => {
+          sumOfFees += tx.fee;
+        });
+        const avarage = sumOfFees / transactions.length;
```
michael1011: Little typo

```suggestion
        const average = sumOfFees / transactions.length;
```

michael1011: We don't need the average fee paid by a transaction in satoshis but the ratio `satoshis / vbyte`

ImmanuelSegol:
@michael1011 Have a look at the `TODO` and tell me what you think.

lib/chain/ChainClient.ts
```diff
@@ -107,6 +112,58 @@ class ChainClient extends BaseClient implements ChainClientInterface {
...
+    let vbytes = 0;
+    transactions.forEach((tx: any) => {
+      vbytes += estimateFee(1, tx.vin, tx.vout);
+      sumOfFees += tx.fee;
+    });
+    return sumOfFees / vbytes;
+  }
+
+  private memPoolAverage = async (): Promise<number> => {
+    return new Promise(async (res, rej) => {
```
michael1011: I don't think wrapping this in a Promise is necessary.

lib/chain/ChainClient.ts
```diff
@@ -107,6 +112,58 @@ class ChainClient extends BaseClient implements ChainClientInterface {
...
+        const average = this.transactionAverage(Object.values(mempool));
+        res(average);
+      } catch (error) {
+        rej(error);
+      }
+    });
+  }
+
+  private bestBlockAverage = (): Promise<number> => {
+    return new Promise(async (res, rej) => {
```
michael1011: Same here for the promise.

armurbalda:
Stale, closing

PR #60, opened 2019-01-08
feat: emit event when invoice not settled

### LINKED ISSUE #46: Add event when paying invoice fails
The gRPC stream `SubscribeInvoices` just shows events about invoices that got paid. But it should also forward data about invoices that couldn't be paid for some reason so that the middleware can tell the frontend that the Swap failed.

ImmanuelSegol:
closes #46


lib/lightning/LndClient.ts
```diff
@@ -310,11 +313,15 @@ class LndClient extends BaseClient implements LightningClient {
...
       .on('data', (invoice: lndrpc.Invoice) => {
+        const rHash = Buffer.from(invoice.getRHash_asB64(), 'base64');
         if (invoice.getSettled()) {
-          const rHash = Buffer.from(invoice.getRHash_asB64(), 'base64');
 
           this.logger.silly(`${this.symbol} LND invoice settled: ${getHexString(rHash)}`);
           this.emit('invoice.settled', rHash);
+        } else {
+
+          this.logger.silly(`${this.symbol} LND invoice failed: ${getHexString(rHash)}`);
```
michael1011: I would rather say:

```suggestion
          this.logger.silly(`Failed to pay ${this.symbol} LND invoice: ${getHexString(rHash)}`);
```

lib/lightning/LndClient.ts
```diff
@@ -310,11 +313,15 @@ class LndClient extends BaseClient implements LightningClient {
...
+          this.emit('invoice.failed', rHash);
```
michael1011: Just because an invoice isn't settled that doesn't mean that Boltz couldn't pay it. The invoice subscription is just about invoices you generate and those invoices getting paid.

I would emit this `invoice.failed` event if the `payInvoice` method fails.

lib/lightning/LndClient.ts
```diff
@@ -309,11 +317,16 @@ class LndClient extends BaseClient implements LightningClient {
...
       .on('data', (invoice: lndrpc.Invoice) => {
+        const rHash = Buffer.from(invoice.getRHash_asB64(), 'base64');
         if (invoice.getSettled()) {
           const paymentReq = invoice.getPaymentRequest();
 
           this.logger.silly(`${this.symbol} LND invoice settled: ${paymentReq}`);
           this.emit('invoice.settled', paymentReq, invoice.getRPreimage_asB64());
+        } else {
+
+          this.logger.silly(`Failed to pay ${this.symbol} LND invoice: ${rHash}`);
```
michael1011: This logic should be removed since it has nothing to do with failing to pay invoices

lib/lightning/LndClient.ts
```diff
@@ -42,6 +42,8 @@ interface GrpcResponse {
 interface LndClient {
   on(event: 'invoice.paid', listener: (invoice: string) => void): this;
   emit(event: 'invoice.paid', invoice: string): boolean;
+  on(event: 'invoice.failed', listener: (rHash: Buffer) => void): this;
```
michael1011: Why are using the `rHash` instead of the invoice itself?

lib/lightning/LndClient.ts
```diff
@@ -217,14 +219,20 @@ class LndClient extends BaseClient implements LightningClient {
...
   public payInvoice = async (invoice: string): Promise<lndrpc.SendResponse.AsObject> => {
-    const request = new lndrpc.SendRequest();
-    request.setPaymentRequest(invoice);
-
-    const repsonse = this.unaryCall<lndrpc.SendRequest, lndrpc.SendResponse.AsObject>('sendPaymentSync', request);
-
-    this.emit('invoice.paid', invoice);
-
-    return repsonse;
+    return new Promise(async (resolve, reject) => {
```
michael1011: Why did you wrap this logic into a `Promise`?

lib/lightning/LndClient.ts
```diff
@@ -40,6 +40,8 @@ interface GrpcResponse {
 interface LndClient {
   on(event: 'invoice.paid', listener: (invoice: string) => void): this;
   emit(event: 'invoice.paid', invoice: string): boolean;
+  on(event: 'invoice.failed', listener: (rHash: Buffer) => void): this;
```
michael1011: This should also have the arguments ` invoice: string` like the `emit` of the `invoice.failed` event
```suggestion
  on(event: 'invoice.failed', listener: (invoice: string) => void): this;
```

lib/lightning/LndClient.ts
```diff
@@ -215,15 +217,18 @@ class LndClient extends BaseClient implements LightningClient {
...
-
-    this.emit('invoice.paid', invoice);
-
-    return repsonse;
+    try {
+      const response = await this.unaryCall<lndrpc.SendRequest, lndrpc.SendResponse.AsObject>('sendPaymentSync', request);
+      this.emit('invoice.paid', invoice);
+      return response;
+    } catch (error) {
+      this.logger.error(`Faild to pay invoice ${request.getPaymentHashString()}: ${error}`);
```
michael1011:
```suggestion
      this.logger.warn(`Failed to pay invoice ${request.getPaymentHashString()}: ${error}`);
```

lib/lightning/LndClient.ts
```diff
@@ -215,17 +217,20 @@ class LndClient extends BaseClient implements LightningClient {
...
-
-    if (response.paymentError === '') {
-      this.emit('invoice.paid', invoice);
+    try {
+      const response = await this.unaryCall<lndrpc.SendRequest, lndrpc.SendResponse.AsObject>('sendPaymentSync', request);
+      if (response.paymentError === '') {
+        this.emit('invoice.paid', invoice);
+      }
+      return response;
+    } catch (error) {
```
michael1011: After doing a little testing it never happened to me that the `this.unaryCall` method threw an error. The `invoice.failed` event should be emitted if `response.paymentError` is not an empty string.

lib/lightning/LndClient.ts
```diff
@@ -215,17 +217,20 @@ class LndClient extends BaseClient implements LightningClient {
...
+      this.logger.warn(`Failed to pay invoice ${request.getPaymentHashString()}: ${error}`);
```
michael1011: I suppose that `request.getPaymentHashString()` wouldn't contain a value since it is not set above and because the request object doesn't get altered when calling the RPC of LND.

lib/lightning/LndClient.ts
```diff
@@ -215,17 +217,22 @@ class LndClient extends BaseClient implements LightningClient {
...
+      const response = await this.unaryCall<lndrpc.SendRequest, lndrpc.SendResponse.AsObject>('sendPaymentSync', request);
+      if (response.paymentError === '') {
+        this.emit('invoice.paid', invoice);
+      } else {
+        this.emit('invoice.failed', invoice);
+      }
+      return response;
+    } catch (error) {
+      const { paymentHash } = await this.decodePayReq(invoice);
+      this.logger.warn(`Failed to pay invoice ${paymentHash}: ${error}`);
```
michael1011: I am not sure why you want to log the hash of the preimage instead of the invoice itself?

PR #61, opened 2019-01-08
refactor: Custom fee for sendToAddress method of Wallet closes #54

### LINKED ISSUE #54: Custom fee for sendToAddress method of Wallet
The wallet should not only use it's default fee but also accept the fee per vbyte for the transaction as parameter.

ImmanuelSegol:
closes #54

### DIFF
--- a/lib/wallet/Wallet.ts
+++ b/lib/wallet/Wallet.ts
@@ -226,7 +226,6 @@ class Wallet {
     };
   }
 
-  // TODO: custom fee
   // TODO: avoid dust to change addresses
   /** Sends a specific amount of funds to and address
    *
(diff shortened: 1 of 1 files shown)


lib/wallet/Wallet.ts
```diff
@@ -232,11 +232,10 @@ class Wallet {
    *
    * @returns the transaction itself and the vout of the provided address
    */
-  public sendToAddress = async (address: string, type: OutputType, isScriptHash: boolean, amount: number):
+  public sendToAddress = async (address: string, type: OutputType, isScriptHash: boolean, amount: number, feePerByte = 10):
```
michael1011: Could you remove the TODO in line [229](https://github.com/BoltzExchange/boltz-backend/pull/61/files#diff-d3a51cad72231c97867cb6412f704432R229)?

PR #68, opened 2019-01-18
Service types

### LINKED ISSUE #49: Invalid argument errors in Service
`Service.ts` should throw errors if an argument has the wrong type or is obviously invalid (negative rate, ...)

ImmanuelSegol:
closes #49


lib/service/Service.ts
```diff
@@ -41,7 +41,34 @@ interface Service {
...
+    if (address === '') throw Errors.INVALID_ARGUMENT('address must be specified');
+  },
+  HAS_INVOICE: ({ invoice }: { invoice: string }) => {
+    if (invoice === '') throw Errors.INVALID_ARGUMENT('invoice must be specified');
+    if (invoice.slice(0, 2) !== 'ln') throw Errors.INVALID_ARGUMENT('invoice is not valid');
+  },
+  HAS_PUBKEY: (pubkey) => {
+    if (pubkey === '') throw Errors.INVALID_ARGUMENT('public key must be specified');
+  },
+  RATE_NOT_VALID: ({ rate }: {rate: number}) => {
```
michael1011: This method should be called `RATE_VALID` for the sake of consistency

lib/service/Service.ts
```diff
@@ -41,7 +41,34 @@ interface Service {
...
 }
 
-// TODO: "invalid argument" errors
+const argChecks = {
+  VALID_CURRENCY: (currency: string) => {
+    if (currency.length >= 3 && currency.length <= 5 && !currency.match(/^[A-Z]+$/))  {
+      throw Errors.INVALID_ARGUMENT('currency must consist of 2 to 5 upper case English letters or numbers');
+    }
+  },
+  HAS_TXID: ({ transactionHash }: {transactionHash: string}) => {
```
michael1011: This should be called `HAS_TXHASH`

lib/service/Service.ts
```diff
@@ -41,7 +41,34 @@ interface Service {
   emit(event: 'invoice.settled', string: string, preimage: string): boolean;
 }
 
-// TODO: "invalid argument" errors
+const argChecks = {
+  VALID_CURRENCY: (currency: string) => {
```
michael1011: Why don't you use destructuring for all parameters so that all of this `argChecks` methods can be called with `argChecks.<method>(args)`?

lib/service/Service.ts
```diff
@@ -41,6 +41,42 @@ interface Service {
   emit(event: 'invoice.settled', string: string, preimage: string): boolean;
 }
 
+const argChecks = {
+  VALID_CURRENCY: ({ currency }: { currency: string }) => {
+    if (currency.length >= 3 && currency.length <= 5 && !currency.match(/^[A-Z]+$/))  {
```
michael1011: Please have a look at this comment: https://github.com/BoltzExchange/boltz-backend/pull/68/files#r249298200

lib/Utils.ts
```diff
@@ -182,7 +182,7 @@ export const getOutputType = (type: number) => {
   switch (type) {
     case 0: return OutputType.Bech32;
     case 1: return OutputType.Compatibility;
-    default: return OutputType.Legacy;
+    case 2: return OutputType.Legacy;
```
michael1011: To solve the compilation error throwing and error on `default` *should* work

lib/service/Service.ts
```diff
@@ -41,6 +41,42 @@ interface Service {
...
+      throw Errors.INVALID_ARGUMENT('currency must consist of 2 to 5 upper case English letters');
+    }
+  },
+  VALID_OUTPUTTYPE: ({ outputType }: { outputType: number }) => {
+    if (outputType % 1 !== 0 && outputType < 0 && outputType > 2) {
```
michael1011: This way the error can never be thrown since the `outputType` would have to be smaller than 0 and bigger than 2.

lib/service/Service.ts
```diff
@@ -41,6 +41,42 @@ interface Service {
...
 }
 
+const argChecks = {
+  VALID_CURRENCY: ({ currency }: { currency: string }) => {
+    if (!(currency.length >= 3 && currency.length <= 5) || !currency.match(/^[A-Z]+$/))  {
+      throw Errors.INVALID_ARGUMENT('currency must consist of 2 to 5 upper case English letters');
+    }
+  },
+  VALID_OUTPUTTYPE: ({ outputType }: { outputType: number }) => {
+    if (outputType % 1 !== 0 && outputType >= 0 && outputType <= 2) {
```
michael1011: With this statement you expressing that: if the number is not divisible by 1 **and** bigger or equal 0 **and** smaller or equal 2 throw the error.

Neither numbers below 0 nor those above 2 would be catched by this if clause although they have no corresponding type and are therefore invalid.

ImmanuelSegol: yeah didn't added a ! now it works

michael1011:
Good to merge once the conflicts are resolved

PR #77, opened 2019-02-15
feat: integrate circleci

ImmanuelSegol:


michael1011:
I managed to install docker with `setup_remote_docker` but for some reason the image isn't starting properly. Maybe that's because the [system that is used for `setup_remote_docker`](https://circleci.com/docs/2.0/building-docker-images/#specifications) is using a hard drive an not a SSD and therefore the startup takes longer and we have to wait until the nodes are able to accept connections.

michael1011:
Should be good now. Please have a look @ImmanuelSegol @peartobear

PR #118, opened 2019-05-03
feat: accept 0-conf for non RBF swap outputs

michael1011:
This PR prepares the backend for accepting 0-conf swaps. In order to accept 0-conf transactions in a safely manner, one has to verify that the transaction does not signal [BIP-125 RBF](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki). I added [that logic to the `SwapNursery`](https://github.com/BoltzExchange/boltz-backend/compare/zero-conf?expand=1#diff-7399a7f20cd561c1188b0f153c349e82R194) and added an additional condition to accepting 0-conf transactions: the middleware has to explicitly allow 0-conf for that specific swap when creating it. This allows for accepting 0-conf for low amounts but forcing the user to wait for bigger swaps.


armurbalda:
So afa I understand, the middle ware just sets these thresholds and accepts 0 conf up to this. No frontend/user involvement, correct (this wouldn't be possible since boltz can't know if it's a RBF tx or not beforehand)?

PR #119, opened 2019-05-07
chore: update Docker images

michael1011:
This PR updates the alpine base images of our Docker images to `3.9.3` and:

- updates Bitcoin Core to version `0.18.0`. The RPC call `generate` was deprecated in that version but it can still be used until `0.19.0` (I had to manually allow that call [in the config](https://github.com/BoltzExchange/boltz-backend/compare/update-bitcoin-core?expand=1#diff-9b9353917242ad1d1b179693e2624521R13)). [Here](https://bitcoincore.org/en/releases/0.18.0/) is the complete changelog for reference.
- updates Litecoin Core to `0.17.1`. Mining a new block with `generate` still fails on Litecoin occasionally. I tried changing the base images to [`debian-slim`](https://hub.docker.com/_/debian) but it didn't have any influence on that issue. There is a PR that fixes this bug already but it will probably take a little while to be merged: https://github.com/litecoin-project/litecoin/pull/585

### DIFF
--- a/docker/berkeley-db/Dockerfile
+++ b/docker/berkeley-db/Dockerfile
@@ -1,8 +1,7 @@
 # Build BerkeleyDB
-FROM alpine:3.9 as berkeley-db
+FROM alpine:3.9.3 as berkeley-db
 
-RUN apk update
-RUN apk upgrade
+RUN apk update && apk upgrade
 RUN apk --no-cache add \
   autoconf \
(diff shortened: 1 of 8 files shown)


michael1011:
I just pushed all of the images to Docker Hub. There are no breaking changes so everything should continue working as it is supposed to.

Issue #137, opened 2019-03-26
bot features


armurbalda:
- [x] `getaddress btc`, `getaddress ltc`
- [x] remote/local channel balance exhausted alerts. threshold configurable.
- [x] `withdraw x.x btc`, `withdraw x.x ltc` (fixed address) **OTP required**
- [x] high number of failed swaps from same IP - alert
- [x] `newaddress btc`, `newadress ltc` (includes above commands listing all addresses, indicating which ones were `(used)`. Sorted by top newest to bottom oldest.
- [x] `getlockedfunds btc`, `getlockedfunds ltc` 
- [x] `getpendingswaps` 
- [ ] `htlctimelockconfig <pairid> <timeout in minutes>`  **OTP required**
- [ ] `gethtlctimelockconfig`

All these commands should also have a CLI implementation, to avoid Discord as a single point of failure.

peartobear:
Added `getlockedfunds btc` and `getlockedfunds ltc` which can help with querying the locked balances and toggle reverse swap on and off manually or automatically directly from the bot, last resort in case of reverse spam attack.

michael1011:
Withdraw command was added by #142

peartobear:
Added `getpendingswaps` command for better swap debugging

peartobear:
Added `gethtlctimelockconfig` `htlctimelockconfig` and OTP requirements along with the requirement for adjacent CLI command to avoid a single point of failure (discord)

PR #142, opened 2019-08-31
feat: add withdraw Discord command

michael1011:
Related to #137 

This PR add a `withdraw` command to the Discord bot. That command is secured by OTP and can be used like this:
- `withdraw <OTP token> <currency> <invoice>`: to withdraw lightning funds
- `withdraw <OTP token> <currency> <address> <amount in whole coins *or* "all">`: to withdraw onchain coins

With this new one we have 9 Discord commands now and I think I should improve the `help` command so that it actually shows how to use the commands instead of just listing them.


michael1011:
I created an issue for the improved help command: #147

PR #183, opened 2020-01-22
fix: add scanned block height to GetInfo

michael1011:
I have observed some weird behavior with Submarine Swaps: when the 0-conf transaction is rejected and the daemon is waiting for the it to be included in a block, it never detects the transaction to be confirmed even if it is mined.

To debug this I added the highest scanned block of the `ZmqClient` (the class handling new blocks and transactions from Bitcoin/Litecoin Core) to the response of the `GetInfo` gRPC call.


armurbalda:
Happening on BTC & LTC?

PR #206, opened 2020-06-17
feat: make invoice payments of prepay minerfee procotol atomic

### LINKED ISSUE #202: prepay miner fee should be made atomic
Currently the server accepts the miner fee payment regardless of whether it will receive the main payment. This leads to issues if no route is found for the main payment.

I believe the flow should be changed as follows:
1. client sends HTLCs for both invoices (`minerFeeInvoice` and `invoice`)
2. Server waits until he has received both HTLCs.
3. Once both HTLCs are received, the server fulfills the HTLC of `minerFeeInvoice` and it sends the onchain funds to the locking address. 
4. If (3.) is not satisfied and invoices time out, the server fails both HTLCs.
5. The rest of the flow is unchanged.

michael1011:
Closes #202 

This PR makes the two invoices of Reverse Swaps in the prepay minerfee protocol *"atomic"* as described in #202. Either both reach the accepted state (pending HTLC) on the LND node Boltz is connected to, or both get cancelled. At least that's the intended behavior. The prepay miner fee is still custodial and a malicious Boltz instance could still claim it before there is a pending HTLC for the bigger invoice.

### DIFF
--- a/lib/db/models/ReverseSwap.ts
+++ b/lib/db/models/ReverseSwap.ts
@@ -17,10 +17,11 @@ type ReverseSwapType = {
   status: string;
   timeoutBlockHeight: number;
 
-  invoice: string;
-  minerFeeInvoice?: string;
-
   preimage?: string;
+  invoice: string;
+
+  minerFeePreimage?: string;
+  minerFeeInvoice?: string;
 
   onchainAmount: number;
   transactionId?: string;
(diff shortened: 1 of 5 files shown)


michael1011:
> In son request body should we include prepay_miner_fee as true or false? Flase if no flag is provided.

The main reason for this prepay miner fee protocol is limiting the exposure of the backend to users paying the hold invoice of Reverse Swaps but not claiming the funds that were sent after the pending HTLC arrived at the Boltz node, which in turn requires the backend to pay onchain fees twice (lockup and refund) and the user getting back all of their offchain funds.

There is no reason for a user *wanting* to use the prepay minerfee protocol since there is only downside for them, so it has to be enforced by the backend. In the current implementation it is a simple boolean in the config that either enables or disables the prepay miner fee protocol.

michael1011:
#238 implements this feature

ecdsa:
@michael1011 AFAICT, #238 does not implement this feature for Bitcoin

PR #296, opened 2022-09-28
fix: missing -f parameter to docker build

patogallaiovlabs:
In a [previous commit](https://github.com/BoltzExchange/boltz-backend/commit/d5f27f98a918c45593c455da386ab9d013a9ea88) the `Dockerfile` was updated and moved to a new folder, but the script in the `package.json` file was not updated.


michael1011:
Nice catch! Thank you!

PR #299, opened 2022-10-27
feat: set max payment fee ration for LND

yuyaogawa:


lib/lightning/LndClient.ts
```diff
@@ -92,7 +93,9 @@ class LndClient extends BaseClient implements LndClient {
   ) {
     super();
 
-    const { host, port, certpath, macaroonpath } = config;
+    const { host, port, certpath, macaroonpath, maxPaymentFeeRatio } = config;
+
+    LndClient.maxPaymentFeeRatio = maxPaymentFeeRatio > 0 ? maxPaymentFeeRatio: 0.03;
```
michael1011, line 98: Because the `maxPaymentFeeRatio` is now set from the constructor, it shouldn't be static anymore.

Else the last client initialized sets the value for all of the `LndClient`s.

michael1011 (changes requested):
One small problem to be fixed, else the PR looks good!

Thank you!

lib/lightning/LndClient.ts
yuyaogawa: Fixed the `maxPaymentFeeRatio` property.

michael1011 (approved):

PR #303, opened 2023-01-18
feat: Swap-in Fee

yuyaogawa:
I added an option to set swap-in fee so that swap in/out fee can be different.


docs/requirements.txt
```diff
@@ -0,0 +1,2 @@
+jinja2<3.1.0
+Markdown<3.2
```
michael1011, line 2: Could you add those dependencies here: https://github.com/BoltzExchange/boltz-backend/blob/master/tools/pyproject.toml
We are using [Poetry](https://python-poetry.org/) for Python dependency management

lib/service/Service.ts
```diff
@@ -674,9 +674,13 @@ class Service {
 
     const rate = getRate(swap.rate!, swap.orderSide, false);
 
-    const percentageFee = this.rateProvider.feeProvider.getPercentageFee(swap.pair);
+    let percentageFee = this.rateProvider.feeProvider.getPercentageFee(swap.pair);
+    const percentageSwapInFee = this.rateProvider.feeProvider.getPercentageSwapInFee(swap.pair);
     const baseFee = this.rateProvider.feeProvider.getBaseFee(onchainCurrency, BaseFeeType.NormalClaim);
 
+    if (percentageSwapInFee !== 0) {
+      percentageFee = percentageSwapInFee;
```
michael1011, line 682: How about we refactor `getPercentageSwapInFee` to have a boolean like `isReverse` as parameter and abstract this into the `FeeProvider`?

lib/service/Service.ts
```diff
@@ -730,22 +734,27 @@ class Service {
...
       rate,
       swap.orderSide,
       invoiceAmount,
       BaseFeeType.NormalClaim,
     );
-    const expectedAmount = Math.floor(invoiceAmount * rate) + baseFee + percentageFee;
+
+    let serviceFee = percentageFee;
+    if (percentageSwapInFee !== 0) {
+      serviceFee = percentageSwapInFee;
```
michael1011, line 747: Same thing here

test/unit/service/Service.spec.ts
```diff
@@ -965,6 +970,7 @@ describe('Service', () => {
     })).rejects.toEqual(Errors.SWAP_WITH_PREIMAGE_EXISTS());
   });
 
+  // TODO: Add anohter test for swapInFee
```
michael1011, line 973: Those tests would be nice to have

michael1011 (changes requested):

PR #396, opened 2023-10-07
feat: CLN reconnect plugin

michael1011:


kilrau:
Let's see how much https://github.com/ElementsProject/lightning/pull/6822 fixes which, according to Rusty, should fix https://github.com/ElementsProject/lightning/issues/5523

michael1011:
Yes, seems to be fixed

PR #402, opened 2023-10-09
refactor: getstats command

michael1011:
- shows totals and grouped by pair
- only includes swaps in which money was sent in failure rates
- only current and last 6 months by default


kilrau:
Concept ACK

PR #414, opened 2023-10-22
fix: hold invoice overpayment

### LINKED ISSUE #413: CLN Hold Invoice Plugin: Handle Overpayments
Currently we are processing swaps as usual when users overpay and there is no limit on how much they can overpay. Onchain (Normal Swaps) there is nothing we can do and rely on the user contacting us with swap details for a refund but for Lightning (Reverse Swaps) we should adhere to [BOLT04](https://github.com/lightning/bolts/blob/master/04-onion-routing.md) behavior:
```
if the amount paid is more than twice the amount expected:
SHOULD fail the HTLC.
SHOULD return an incorrect_or_unknown_payment_details error.
Note: this allows the origin node to reduce information leakage by altering the amount while not allowing for accidental gross overpayment.
```
- immediately fail incoming  payments for a hold invoice that already has a pending payment and
- immediately fail payments that overpay 2x or more as per above `BOLT04` definition.

LND's integrated holdinvoice implemenation already behaves like this. Our own CLN hold invoice plugin should too.

michael1011:
Closes  #413


kilrau:
:fire:

kilrau (approved):
Tested
- 3x overpayment - rejected
- simultaneously from two different wallets - one went through, one rejected

LGTM

PR #425, opened 2023-10-31
feat: PostgreSQL as database

michael1011:


michael1011:
For reference, the database migration script for pgloader:

```
load database 
     from sqlite:///home/michael/boltz.db
     into pgsql://boltz:boltz@localhost/boltz

 with quote identifiers, data only, reset sequences

  set work_mem to '16MB', maintenance_work_mem to '512 MB';
```

PR #430, opened 2023-11-06
feat: api landing page favico, misc fixes

kilrau:
Closes https://github.com/BoltzExchange/api-landing/issues/2


lib/api/static/api.html
```diff
@@ -41,10 +42,10 @@
...
-  <h1>Privacy First, Non-Custodial Bitcoin Exchange</h1>
-  <p>Can't be evil. You are in control of your money. Always.</p>
+  <h1>API Endpoint</h1>
```
michael1011, line 45: I don't like this heading. Looks a little out of place

lib/api/static/api.html
```diff
@@ -41,10 +42,10 @@
...
+  <a class="btn" target="_blank" href="https://docs.boltz.exchange/v/api/">API Docs</a>
+  <a class="btn" target="_blank" href="https://api.boltz.exchange/getpairs">API Test</a>
```
michael1011, line 47: It's also in the docs and seems redundant here

michael1011 (changes requested):

lib/api/static/api.html
kilrau: Heading as in *different heading style* or *wording*?

lib/api/static/api.html
kilrau: The `API Test` button? I knew you wouldn't. Removed.

PR #463, opened 2024-01-15
docs: redeem script clarification

kilrau:


docs/api.md
```diff
@@ -207,7 +207,7 @@ For UTXO chains, `/createswap` requests have to contain one additional parameter
 
 Responses also contain one additional value:
 
-* `redeemScript`: redeem script from which the `address` is derived. The redeem script can (and should!) be used to verify that Boltz didn't try to cheat by providing an address without HTLC.
+* `redeemScript`: redeem script from which the `address` is derived. The redeem script should be used, [to verify](dont-trust.-verify.md#utxo-chain-address-verification) that Boltz didn't provide a different address.
```
michael1011, line 210: Different to what?
```suggestion
* `redeemScript`: redeem script from which the `address` is derived. The redeem script should be used, [to verify](dont-trust.-verify.md#utxo-chain-address-verification) that Boltz did provide the correct address.
```

PR #484, opened 2024-02-12
V2 docs

michael1011:


lib/api/v2/WebSocketHandler.ts
```diff
@@ -118,53 +117,63 @@ class WebSocketHandler {
...
+
+        const args = subscribeData.args.map((id) => {
+          const status = this.controller.pendingSwapInfos.get(id);
+          if (status === undefined) {
+            return {
+              id,
+              error: Errors.SWAP_NOT_FOUND(id).message,
+            };
+          }
+
```
jackstar12, line 152: Instead of updating the reverse maps seperately above, we can just do it here
```ts
this.swapToSockets.get(id).add(socket)
this.socketToSwaps.get(socket).add(id)
```

PR #485, opened 2024-02-15
Type fixes

Czino:
I found 2 minor differences between the actual data types used and what has been documented.


michael1011:
Nice catch! Thank you

PR #486, opened 2024-02-15
fix: API v2, wording, style

kilrau:
Preview: https://boltz.gitbook.io/boltz-backend-dev


docs/0-conf.md
```diff
@@ -12,9 +12,11 @@ As a precautionary measure, Boltz enforces a few rules when it comes to 0-conf.
 * In **Normal Submarine Swaps** in which the user sends the chain transaction, _Boltz_ is taking the risk by accepting unconfirmed transactions.
 * In **Reverse Submarine Swaps** where the user receives the chain transaction from Boltz, _the user_ is at risk for accepting the unconfirmed transaction.
 
-_0-conf Swaps are subject to network conditions and generally only available on UTXO chains like Bitcoin._
+_0-conf Swaps are subject to network conditions and generally only available on UTXO chains like Bitcoin or Liquid._
```
michael1011, line 15:
```suggestion
_0-conf Swaps are subject to network conditions and only available on UTXO chains like Bitcoin or Liquid._
```

EVM txs are always fully replaceable

docs/lifecycle.md
```diff
@@ -29,8 +29,8 @@ When a Normal Submarine Swap is created, it passes through the following states:
...
    * `invoice.failedToPay`: if paying the invoice failed. In this case the user needs to broadcast a refund transaction to reclaim the locked up chain bitcoin.
-6. `transaction.claim.pending`: Taproot swaps are not claimed immediately after the invoice has been paid. Boltz either waits for the client to post a cooperative signature for a key path spend or batches script path spends. This status indicates that Boltz is ready for the creation of a cooperative signature
-7. `transaction.claimed`: indicates that after the invoice was successfully paid, the chain Bitcoin were successfully claimed _by Boltz_. This is the final status of a successful Normal Submarine swap.
+6. `transaction.claim.pending`: This status indicates that Boltz is ready for the creation of a cooperative signature for a keypath spend. Taproot Swaps are not claimed immediately by Boltz after the invoice has been paid, but instead Boltz waits for the API client to post a signature for a key path spend, which is the more efficient and private method. If the API client does not cooperate in a key path spend, Boltz includes its claim transaction in a script path spend batch.
```
michael1011, line 32:
```suggestion
6. `transaction.claim.pending`: This status indicates that Boltz is ready for the creation of a cooperative signature for a keypath spend. Taproot Swaps are not claimed immediately by Boltz after the invoice has been paid, but instead Boltz waits for the API client to post a signature for a key path spend. If the API client does not cooperate in a key path spend, Boltz will eventually claim via the script path.
```

The reason is not really relevant for the status and the batching is also not important to the client

docs/claiming-swaps.md
```diff
@@ -5,52 +5,51 @@ description: >-
...
-# 🙋 Claim & Refund Transactions
+# 🙋♂ Claim & Refund Transactions
 
-Boltz API clients need to craft and broadcast in order to:
+Boltz API clients need to craft and broadcast
 
-- **claim transactions** to claim chain bitcoin and successfully complete **Reverse Submarine Swaps**
-- **refund transactions** to refund chain bitcoin to get back coins locked for failed **Submarine Swaps**
+* **claim transactions** to claim chain bitcoin and successfully complete **Reverse Submarine Swaps**
+* **refund transactions** to claim refunds for chain bitcoin locked in failed **Submarine Swaps**
```
michael1011, line 13: Not a fan of the "claim refund" terminology. Kinda confusing to use the word "claim" for refunds

docs/claiming-swaps.md
```diff
@@ -5,52 +5,51 @@ description: >-
...
 
 ```
 OP_1
 <tweaked aggregated public key>
 ```
 
-Key path spends should always be preferred over using a script path spends.
-Not only, to save on miner fees but also, because key path spends are better for privacy.
-Partial signatures from Boltz are always using `SIGHASH_DEFAULT`.
+Key path spends should be preferred over script path spends. One reason is the smaller chain footprint and thus cheaper miner fees and another reason is because key path spends are better for privacy as they don't reveal anything about the atomic swap on the chain.&#x20;
```
michael1011, line 28:
```suggestion
Key path spends should be preferred over script path spends. One reason is the smaller chain footprint and thus cheaper miner fees and another reason is because key path spends are better for privacy as they don't reveal anything about the atomic swap on the chain.
```

docs/claiming-swaps.md
```diff
@@ -5,52 +5,51 @@ description: >-
...
 #### Claim
 
-To allow Boltz to claim Submarine Swaps cooperative, `GET /swap/submarine/{id}/claim` to obtain the necessary information to create a partial signature.
-Give Boltz your partial signature with `POST /swap/submarine/{id}/claim` and it will broadcast the key path spend claim transaction.
+In order for Boltz to claim Submarine Swaps cooperatively, use `GET /swap/submarine/{id}/claim`  to obtain the necessary information to create a partial signature. Provide your partial signature to Boltz with `POST /swap/submarine/{id}/claim` and Boltz will broadcast the key path spend claim transaction.
 
 In case the client does not cooperate, Boltz will eventually broadcast a script path claim transaction to sweep the UTXO.
 
+{% hint style="info" %}
+Partial signatures from Boltz use `SIGHASH_DEFAULT`.
```
michael1011, line 41: That's also relevant for refunds

michael1011 (changes requested):

docs/claiming-swaps.md
kilrau: We had this topic before. We'll keep it for now.

docs/claiming-swaps.md
kilrau: Moved into Taproot section

PR #500, opened 2024-03-08
feat: add setswapstatus to boltz-cli

### LINKED ISSUE #495: Update swap status CLI tool
Parameters:
- swap ID
- status
                          
Use case A:
Mark the swap status as `failedToPay` so that the user is able to refund immediately.

maybeast:
This commit adds a new `boltz-cli setswapstatus <swapID> <status>` command to update swap statuses manually.

Closes #495


lib/service/Service.ts
```diff
@@ -864,6 +865,25 @@ class Service {
...
+    if (!isSwapUpdateEvent(status)) {
+      throw Errors.INVALID_SWAP_UPDATE_EVENT(status);
+    }
+    const swap = await SwapRepository.getSwap({
+      id: id,
+    });
+    if (!swap) {
+      throw Errors.SWAP_NOT_FOUND(id);
+    }
+    await SwapRepository.setSwapStatus(swap, status);
```
michael1011, line 884: 1) does not handle reverse swaps
2) just updating the status in the database does not update the cache in `Controller.ts`

Having that cache in there and the way it is updated is not optimal, might want to consider a refactor to be able to update the cached status properly via gRPC

michael1011 (changes requested):

maybeast:
Updated. As discussed internally we're only going to handle 2 statuses, `invoice.pending` and `invoice.failedToPay`, within the scope of this PR until we refactor the swap update event handling.

test/unit/service/Service.spec.ts
```diff
@@ -2489,4 +2492,55 @@ describe('Service', () => {
...
+      const eventName = SwapUpdateEvent.TransactionFailed;
+
+      try {
+        await setSwapStatus({ id: '1', status: eventName });
+      } catch (error) {
+        const err = error as Error;
+        expect(err.message).toBe(
+          Errors.SWAP_UPDATE_EVENT_NOT_ALLOWED(eventName).message,
+        );
+      }
```
michael1011, line 2511: `await expect(<promise>).rejects.toEqual()`
Is nicer syntax

lib/service/Service.ts
```diff
@@ -1831,5 +1868,7 @@ class Service {
   };
 }
 
+export const cancelledViaCliFailureReason = 'canceled via CLI';
```
michael1011, line 1871: That reason is shown directly to the user. Not sure if we want to show `via CLI`. How about `payment has been cancelled`?

michael1011 (changes requested):
Works! Just some minor feedback on wording and code style

michael1011 (approved):

PR #511, opened 2024-03-20
Chain swaps

michael1011:


Joelokon (approved):

Joelokon (approved):

kilrau:
@Joelokon thanks for the reviews! How'd you become interested in the project?

Joelokon (approved):

Joelokon (approved):

Joelokon (approved):

lib/db/models/ChainSwap.ts
```diff
@@ -0,0 +1,95 @@
...
+  public static load = (sequelize: Sequelize): void => {
+    ChainSwap.init(
+      {
+        id: {
+          type: new DataTypes.STRING(255),
+          primaryKey: true,
+          allowNull: false,
+        },
+        pair: { type: new DataTypes.STRING(255), allowNull: false },
+        orderSide: { type: new DataTypes.INTEGER(), allowNull: false },
```
jackstar12, line 51: Maybe we should try to adopt the from / to currency model? Model could still expose those as wrappers for backwards compatible, but it would greatly simplify the fee / stat related queries.

test/integration/service/cooperative/ChainSwapSigner.spec.ts
```diff
@@ -0,0 +1,759 @@
...
+    test('should throw when signing refund for swap that could not be found', async () => {
+      ChainSwapRepository.getChainSwap = jest.fn().mockResolvedValue(undefined);
+      const id = 'swap';
+
+      await expect(
+        signer.signRefund(id, Buffer.alloc(0), Buffer.alloc(0), 0),
+      ).rejects.toEqual(Errors.SWAP_NOT_FOUND(id));
+    });
+
+    test('should throw when signing refund on UTXO based chain', async () => {
```
maybeast, line 324: not UTXO based

lib/swap/UtxoNursery.ts
```diff
@@ -92,139 +122,209 @@ class UtxoNursery extends TypedEventEmitter<{
...
+        },
+      );
 
-        if (!swap) {
-          continue;
-        }
+      if (
+        !swap ||
+        wallet.symbol !== swap.receivingData.symbol ||
+        !wallet.decodeAddress(swap.receivingData.lockupAddress).equals(script)
```
jackstar12, line 182: Isnt the check against the `wallet.symbol` enough?

lib/swap/UtxoNursery.ts
```diff
@@ -92,139 +122,209 @@ class UtxoNursery extends TypedEventEmitter<{
...
+      preimage: detectPreimage(inputIndex, transaction),
+    });
+  };
+
+  private checkChainSwapClaim = async (
+    chainClient: IChainClient,
+    transaction: Transaction | LiquidTransaction,
+    inputIndex: number,
+    inputTransactionId: string,
+    input: { hash: Buffer; index: number },
```
jackstar12, line 276: why 3 params? wouldnt the `input` one be enough? same for `checkReverseSwapClaims`

lib/swap/SwapNursery.ts
```diff
@@ -981,55 +1215,81 @@ class SwapNursery extends TypedEventEmitter<SwapNurseryEvents> {
...
+        ? await ReverseSwapRepository.setReverseSwapStatus(
+            swap as ReverseSwap,
+            SwapUpdateEvent.TransactionFailed,
+            Errors.COINS_COULD_NOT_BE_SENT().message,
+          )
+        : await ChainSwapRepository.setSwapStatus(
+            swap as ChainSwapInfo,
+            SwapUpdateEvent.TransactionFailed,
+            Errors.COINS_COULD_NOT_BE_SENT().message,
+          ),
```
jackstar12, line 1270: might want to add wrappers for `setSwapStatus` and `setTransactionRefunded`

jackstar12 (approved):

Joelokon (approved):

kilrau:
Gave this another run on regtest. Observations/questions:
- How do I configure a separate fee for chain swaps via `boltz.conf`? Currently it seems to be taking 0.25% from Liquid Reverse Swaps for both sides. I know this is because of our legacy and it doesn't have to be nice, but we need to be able to set separate fees for all chain swaps *per direction* too.
- I adjusted `boltz.conf` to match mainnet values, but whatever I do min amount for chain swaps seems to be hardcoded to `102000` sats - what's that about?
- Network fee stays the same when I switch directions, this definitely shouldn't be
[Screencast from 2024-05-13 15-01-17.webm](https://github.com/BoltzExchange/boltz-web-app/assets/19181985/7b8fd160-5c7d-4b13-871b-21cd9fd89572)

lib/db/models/ChainSwap.ts
michael1011: I am thinking to do this in a followup PR so that the scope of this one doesn't get out of hand

lib/swap/SwapNursery.ts
```diff
@@ -85,9 +102,12 @@ class SwapNursery extends TypedEventEmitter<SwapNurseryEvents> {
   private currencies = new Map<string, Currency>();
 
   // Locks
-  public readonly lock = new AsyncLock();
+  public readonly lock = new AsyncLock({
+    maxPending: 10_000,
+  });
```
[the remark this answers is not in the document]
michael1011: Just a bump from the default 1000

lib/service/Utils.ts
```diff
@@ -0,0 +1,25 @@
...
+) => {
+  const currency = currencies.get(symbol);
+  if (!currency) {
+    throw Errors.CURRENCY_NOT_FOUND(symbol);
+  }
+
+  return currency;
+};
+
+export const calculateTimeoutDate = (chain: string, blocksMissing: number) => {
```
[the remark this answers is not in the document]
michael1011: It's also used in the Service. And I prefer exported functions over static methods

lib/swap/UtxoNursery.ts
michael1011: `inputIndex` because that's the number of the input in the transaction, the `index` in `input` refers to the output number of the UTXO being spent
`inputTransactionId` is avoid having to reverse and serialize the `hash` in `input` twice. That's a micro optimization, that can be undone for the sake of simplicity in the code

kilrau:
liquid refunds working now :+1:

PR #518, opened 2024-04-02
feat: add getLockedFunds gRPC call

maybeast:
Expose `getLockedFunds` call via gRPC.


lib/service/Service.ts
```diff
@@ -894,6 +896,33 @@ class Service {
...
+      const chainCurrency = getChainCurrency(
+        pair.base,
+        pair.quote,
+        pending.orderSide,
+        true,
+      );
+
+      const chainArray = acc.get(chainCurrency) || [];
+      chainArray.push(pending);
+      acc.set(chainCurrency, chainArray);
```
michael1011, line 920: How about using `DefaultMap` instead?

proto/boltzrpc.proto
```diff
@@ -212,3 +215,17 @@ message DevHeapDumpRequest {
 }
 
 message DevHeapDumpResponse {}
+
+message ReverseSwap {
```
michael1011, line 219:
```suggestion
message LockedFunds {
```

proto/boltzrpc.proto
```diff
@@ -212,3 +215,17 @@ message DevHeapDumpRequest {
...
+  string id = 1;
```
michael1011, line 220:
```suggestion
  string swap_id = 1;
```

test/unit/notifications/CommandHandler.spec.ts
```diff
@@ -141,6 +147,7 @@ jest.mock('../../../lib/service/Service', () => {
 
         return res;
       },
+      getLockedFunds: async () => mockedLockedFunds,
```
michael1011, line 150:
```suggestion
      getLockedFunds: jest.fn().mockResolvedValue(mockedLockedFunds),
```

michael1011 (changes requested):

PR #535, opened 2024-04-25
fix: make mpay not crash on rpc if you pass pay compatible args

dni:
also throw the exception of mpay so the rpc client can throw aswell


kilrau:
Are you fixing this one up as discussed? If not please close @dni

PR #544, opened 2024-05-13
chore: minor logging clarification reverse swaps hold invoice payment

kilrau:


michael1011 (changes requested):
There is no `pending` state. LND and our hold invoice plugin both call that state `accepted`

PR #555, opened 2024-06-01
refactor: swap status cache

michael1011:
Fetching the status of all swaps on startup takes a couple seconds and is not necessary. Almost all swaps in the database are already done and their status will never be fetched again.
Therefore, we only fetch the swap status as needed and cache after fetching.

### DIFF
--- a/test/unit/api/SwapInfos.spec.ts
+++ b/test/unit/api/SwapInfos.spec.ts
@@ -60,9 +60,9 @@ describe('SwapInfos', () => {
 
       await expect(swapInfos.get(id)).resolves.toEqual(status);
 
-      expect(SwapRepository.getSwap).toHaveBeenCalledTimes(0);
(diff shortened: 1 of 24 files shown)


test/unit/api/SwapInfos.spec.ts
```diff
@@ -25,318 +30,372 @@ describe('SwapInfos', () => {
...
+      await expect(swapInfos.get(id)).resolves.toEqual(status);
 
-      const channelCreation = {
-        fundingTransactionId: 'txId',
-        fundingTransactionVout: 12,
-      };
-      ChannelCreationRepository.getChannelCreation = jest
-        .fn()
-        .mockResolvedValue(channelCreation);
+      expect(SwapRepository.getSwap).toHaveBeenCalledTimes(0);
```
maybeast, line 63: `not.toHaveBeenCalled();` reads better

maybeast (approved):

PR #564, opened 2024-06-09
Granular reset of mpay memory

michael1011:
This PR adds the ability to reset the memory of mpay more granularly. The permanent memory in the database, the temporary one of recently failed channels or both can be reset.


tools/plugins/mpay/rpc/server.py
```diff
@@ -106,16 +106,24 @@ def ListPayments(  # noqa: N802
 
     def ResetPathMemory(  # noqa: N802
         self,
-        request: ResetPathMemoryRequest,  # noqa: ARG002
+        request: ResetPathMemoryRequest,
         context: grpc.ServicerContext,  # noqa: ARG002
     ) -> ResetPathMemoryResponse:
-        res = self._routes.reset()
-        self._mpay.reset_excludes()
+        if not request.exclude_temporary_memory:
```
maybeast, line 112: fwiw, the double negative here `not` + `exclude` is confusing to read, but should work :p

PR #617, opened 2024-07-23
CLN static channels backups to S3

michael1011:


tools/plugins/backup/backup.py
```diff
@@ -99,7 +172,7 @@ def log_could_not_init(msg: str) -> None:
 def backup_upload(plugin: Plugin) -> dict[str, Any]:
     """Upload a backup of the current SCB to the configured WebDAV."""
     try:
-        wd.upload_backup()
+        backup.upload_backup()
```
maybeast, line 175: do we want some kind of retry mechanism in case of failures?

michael1011: Eventually yes, but for now I just want to ditch webdav. Could you create an issue for retries?

maybeast (approved):

PR #619, opened 2024-07-23
chore: use prettier for Markdown files

michael1011:


dni (approved):

kilrau:
Looking good, but gimme a sec to create gitbook preview to make sure nothing is bricked

PR #627, opened 2024-07-28
feat: custom Core RPC wallet names

michael1011:


lib/chain/ChainClient.ts
```diff
@@ -283,7 +283,7 @@ class ChainClient<T extends SomeTransaction = Transaction>
   };
 
   public getWalletInfo = async (): Promise<WalletInfo> => {
```
maybeast, line 285: unused fn?

maybeast (approved):

lib/chain/ChainClient.ts
michael1011: Yes

PR #650, opened 2024-08-18
feat(mpay): add start_id and limit to mpay-list

maybeast:


tools/plugins/mpay/protos/mpay.proto
```diff
@@ -43,6 +43,8 @@ message ListPaymentsRequest {
     string bolt11 = 1;
     string payment_hash = 2;
   }
+  optional uint64 start_id = 3;
+  optional uint32 limit = 4;
```
michael1011, line 47: It should be either `bolt11` *or* `payment_hash` or those two for pagination. Paginating a single entry doesn't make much sense

michael1011 (changes requested):

michael1011 (approved):

PR #656, opened 2024-08-24
refactor: move SSE swap update streams to sidecar

michael1011:


boltzr/src/config.rs
```diff
@@ -257,6 +262,10 @@ erc20SwapAddress = "0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512"
                     request_timeout: None,
                     max_retries: None,
                 }),
+                api: crate::api::Config {
+                    host: "127.0.0.1".to_string(),
+                    port: 9005,
```
maybeast, line 267: needs nginx conf adjustments for regtest later?

maybeast (approved):

boltzr/src/config.rs
michael1011: Yes

Issue #579, opened 2024-06-24
Chain swaps: allow any amount


kilrau:
Currently we allow creation of chain swaps only for a predetermined `userLockAmount` (or alternatively `serverLockAmount`), see https://api.boltz.exchange/swagger#/Chain%20Swap/post_swap_chain , but since we are not committing to a hash of a Lightning invoice with a fixed amount we should be able to accept whatever user is sending and process the swap nevertheless. As long as amount is within our limits.

API change has to be non-breaking.

PR #657, opened 2024-08-26
Arbitrary Chain Swap amounts

michael1011:
Starts with #579


maybeast (approved):

lib/api/v2/routers/SwapRouter.ts
```diff
@@ -1492,6 +1492,93 @@ class SwapRouter extends RouterBase {
...
+    /**
+     * @openapi
+     * components:
+     *   schemas:
+     *     Quote:
+     *       type: object
+     *       properties:
+     *         quote:
+     *           type: number
+     *           description: New quote for a Swap. Amount that the server will lock for Chain Swaps
```
jackstar12, line 1504:
```suggestion
     *         amount:
     *           type: number
     *           description: New amount that the server will lock for Chain Swaps
```

PR #673, opened 2024-10-03
chore: bump max invoice description length

michael1011:


ndeet:
Hey, thanks for fixing. Any reason you not just cut of if the string is longer than the limit. It will be only a matter of time before somebody else has a >100 char string that will fail again?

michael1011:
> Any reason you not just cut off if the string is longer than the limit.

That 100 char limit right now is completely arbitrary and exists to prevent API callers from writing too much data in the database of our lightning nodes. We are trying to figure out reasonable validation for which characters and how many of them should be allowed. Do you have feedback on that?

> It will be only a matter of time before somebody else has a >100 char string that will fail again?

Yes, but I think it's the clients responsibility to handle that and potentially slice the string into an appropriate format

ndeet:
> We are trying to figure out reasonable validation for which characters and how many of them should be allowed. Do you have feedback on that?

Ok got it, by documenting the 100 char limit and just stripping if it is more you would not fill your db and still not make requests fail if they their clients auto generate too long strings like `${'Paid to A Very Uncommon Store Name (Order ID: 6f693cac-20c3-4e3b-a47a-702276fddd8c; Transaction ID: 6f693cac-20c3-4e3b-a47a-702276fddd8c)'}`

But ok, if you cover 99% of cases with 100 chars it is probably fine like this. Just the memo field allows more chars and people will maybe assume that your api to allows the same size as the LN spec and run into this randomly.

michael1011:
> and still not make requests fail if they their clients auto generate too long strings

To be honest I'd rather fail the request than have the client figure out if something and what was stripped

ndeet:
Fair enough

PR #705, opened 2024-11-03
Backups refactor

### LINKED ISSUE #620: Retry SCB backups
Currently, we don't retry uploading SCB backups if it fails.

https://github.com/BoltzExchange/boltz-backend/pull/617#discussion_r1687887693

michael1011:
Closes #620

Moves backups into the sidecar and changes compression algorithms:

- database backups are compressed with zstd instead of gzip which improves the ratio and is faster
- LND SCBs are compressed with gzip (more than good enough here and way easier to use) instead of uploading them raw. The compression saves ~40% space


boltzr/src/backup/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+                    let self_cp = self.clone();
+                    let mut backup_stream = lnd.subscribe_channel_backups();
+
+                    tokio::spawn(async move {
+                        if let Err(err) = self_cp.lnd_backup(&symbol, &mut lnd).await {
+                            error!("LND backup failed: {}", err);
+                        }
+
+                        loop {
+                            while let Ok(backup) = backup_stream.recv().await {
```
jackstar12, line 72: What if the stream dies?

boltzr/src/backup/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct Backup {
+    cancellation_token: CancellationToken,
+
+    interval: String,
+    db_config: crate::db::Config,
+
+    provider: Option<providers::s3::S3>,
```
jackstar12, line 37: Shouldnt this be `providers::BackupProvider`?

boltzr/src/backup/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+        db_config: crate::db::Config,
+    ) -> anyhow::Result<Self> {
+        Ok(Backup {
+            db_config,
+            cancellation_token,
+            interval: config.interval.unwrap_or(DEFAULT_INTERVAL.to_string()),
+            provider: if let Some(config) = config.simple_storage {
+                Some(providers::s3::S3::new(&config).await?)
+            } else {
+                warn!("no backup provider configured");
```
jackstar12, line 53: why not error in this case? this seems like a case of invalid configuration to me

michael1011: Yeah, that would actually make it quite a bit easier to handle

boltzr/src/backup/mod.rs
michael1011: In theory, yes. In practice it's the only one we have implemented in the foreseeable future, so it's fine

jackstar12 (approved):

boltzr/src/backup/mod.rs
```diff
@@ -115,6 +132,44 @@ impl Backup {
         Ok(())
     }
 
+    async fn retry_loop(self) {
+        let dur = Duration::from_secs(SCB_RETRY_INTERVAL_SECONDS);
+        debug!("Retrying failed backups every {:#?}", dur);
+        let mut interval = tokio::time::interval(dur);
+
+        loop {
+            interval.tick().await;
```
jackstar12, line 141: shouldn't this loop respect the `cancellation_token` aswell?

michael1011: It not caring about that wouldn't block shutdown but yes, it should for good practice

boltzr/src/backup/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+        let mut dump_cmd = tokio::process::Command::new("pg_dump")
+            .env("PGPASSWORD", self.db_config.password.clone())
+            .arg("-U")
+            .arg(self.db_config.username.clone())
+            .arg("-h")
+            .arg(self.db_config.host.clone())
+            .arg("-p")
+            .arg(format!("{}", self.db_config.port))
+            .arg("-d")
+            .arg(self.db_config.password.clone())
```
maybeast, line 219: shouldn't this be database?

boltzr/src/main.rs
```diff
@@ -117,8 +119,49 @@ async fn main() {
...
+        config.network,
+        config.currencies,
+        config.liquid,
+    )
+    .await
+    {
+        Ok(currencies) => currencies,
+        Err(err) => {
+            error!("Could not connect to nodes: {}", err);
+            std::process::exit(1);
```
maybeast, line 133: Do we really want backend unrestartable when we can't connect to one of the nodes?

PR #709, opened 2024-11-05
Post release fixes

michael1011:


maybeast (approved):

test/integration/wallet/providers/CoreWalletProvider.spec.ts
```diff
@@ -122,22 +126,36 @@ describe('CoreWalletProvider', () => {
...
 
-  it('should get unconfirmed balance correctly', async () => {
-    await provider.sendToAddress(
-      await provider.getAddress(''),
-      10000,
-      undefined,
-      '',
-    );
+    it('should get safe unconfirmed balance correctly', async () => {
+      await provider.sendToAddress(
```
jackstar12, line 137: So our own utxos are considered save to spend, right?

jackstar12 (approved):

test/integration/wallet/providers/CoreWalletProvider.spec.ts
michael1011: Our own change is (and what we send to ourselves which we don't do in prod), yes. I should probably add a comment to that

PR #715, opened 2024-11-09
EVM docs clarifications

### LINKED ISSUE #446: Docs: Rootstock Addons
- [ ] Still need to add more details on which values are needed for EVM refunds and best also an rpc call example for the emergency procedure how to obtain them via contract logs if refund info was lost. Need @michael1011 's help for that.
- [ ] add rsk testnet and mainnet contract addresses and bytecode verification examples to "Don't trust. Verify!"

michael1011:
Closes #446


boltzr/src/swap/filters.rs
kilrau, line 1: That change supposed to be in here?

docs/claiming-swaps.md
```diff
@@ -135,25 +135,25 @@ structured like this:
...
 from the response of the swap creation should be stored. If this information is
 not available, all parameters required for a refund can also be queried from the
 `Lockup` event logs of the contract. The event logs are indexed by
-`refundAddress`,which is the address with which the client locked the coins.
+`refundAddress`, which is the address with which the client locked the coins.
 
-To refund before the time lock of the swap expired, an EIP-712 signature can be
-requested from Boltz. Use `GET /swap/submarine/{id}/refund` to obtain this
+To refund before the time lock of the swap has expired, an EIP-712 signature can
+be requested from Boltz. Use `GET /swap/submarine/{id}/refund` to get this
```
kilrau, line 156: link swagger here too

lib/swap/NodeSwitch.ts
kilrau, line 1: shouldn't be in here

test/unit/swap/NodeSwitch.spec.ts
kilrau, line 1: shouldn't be in here

PR #721, opened 2024-11-12
feat: Liquid 0-conf checks via API

michael1011:


lib/chain/ElementsWrapper.ts
```diff
@@ -75,29 +80,10 @@ class ElementsWrapper
...
-            testResult['reject-reason'] !== undefined &&
-            !ElementsWrapper.zeroConfCheckAllowedErrors.includes(
-              testResult['reject-reason'],
-            )
-          ) {
-            this.logger.warn(
-              `Rejected ${this.symbol} 0-conf transaction (${transaction.getId()}): ${testResult['reject-reason']}`,
-            );
-            return;
+          if (await this.zeroConfCheck.checkTransaction(transaction)) {
```
maybeast, line 84: Do we want a fallback in case the zeroConfTool is down?

michael1011: The endpoint not working will effectively result in us not accepting 0-conf anymore which is imho fine

maybeast (approved):

jackstar12 (approved):

PR #725, opened 2024-11-17
fix: chunk batch claim transactions

michael1011:
Creating batch claim transactions with more than ~20 inputs on Liquid does not work with Node.js. To work around that, we create multiple batch sweeps with at most 15 inputs each.

### DIFF
--- a/lib/service/cooperative/DeferredClaimer.ts
+++ b/lib/service/cooperative/DeferredClaimer.ts
@@ -9,6 +9,7 @@ import {
 } from '../../Core';
 import Logger from '../../Logger';
 import {
+  arrayToChunks,
   formatError,
   getChainCurrency,
   getHexBuffer,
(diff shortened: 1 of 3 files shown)


maybeast (approved):

lib/service/cooperative/DeferredClaimer.ts
```diff
@@ -238,30 +241,45 @@ class DeferredClaimer extends CoopSignerBase<
...
+      try {
+        claimed = claimed.concat(
+          ...(await this.broadcastClaim(symbol, toClaimChunk)),
+        );
+      } catch (e) {
+        lastError = e;
+        this.logger.error(
+          `Batch claim for currency ${symbol} failed: ${formatError(e)}`,
+        );
+        await this.lock.acquire(DeferredClaimer.swapsToClaimLock, async () => {
```
jackstar12, line 268: Couldn't all of this logic be simplified if the lock was held for the entirety of the function call?

michael1011: Yes, it would be simpler, but I do not want to hold the lock for the entirety of the function call. If some new swap claim is deferred, it could have to wait *a while* until it gets the lock if a batch claim is running at that time.

PR #731, opened 2024-11-30
Fix required properties in Swagger doc

michael1011:


lib/swap/UtxoNursery.ts
```diff
@@ -689,6 +690,13 @@ class UtxoNursery extends TypedEventEmitter<{
           : (swap as ChainSwapInfo).sendingData.lockupAddress,
       ),
     );
+    if (SuccessSwapUpdateEvents.includes(swap.status as SwapUpdateEvent)) {
```
jackstar12, line 693: When is this the case? Clients accepting zero-conf from server?

jackstar12 (approved):

lib/swap/UtxoNursery.ts
michael1011: It's a race condition

PR #734, opened 2024-12-03
feat: batch sweep EVM currencies

michael1011:
Depends on https://github.com/BoltzExchange/boltz-core/pull/127


maybeast (commented):
utACK

Issue #516, opened 2024-03-30
Continue retrying lightning payments when websocket is connected again


kilrau:
The idea is to break our payment retry loops for offline Lightning wallets after X loops and only continue trying for another X loops if the websocket for the swap is active. Signifying the user is online and active.

Even though we scaled down retry loops, we still have cases where users start swaps and just never come online again.

But websocket != Lightning wallet so this is just an idea really.

kilrau:
Closing this in favor of https://github.com/BoltzExchange/boltz-backend/issues/885

PR #740, opened 2024-12-10
feat: abandon swaps that trigger MC reset loop

maybeast:


michael1011 (changes requested):
Imho, abandoning a swap after one reset is too harsh. How about we revisit this after #516 is implemented and only give up on paying until the client signals that it is ready?

PR #762, opened 2024-12-30
feat: check submarine swaps for expired invoices

### LINKED ISSUE #367: Move to swap failed, invoice expired state immediately when invoice expired
Currently one sees this state when on-chain tx is not confirmed yet but invoice already expired:
Screenshot from 2023-07-19 21-07-00
Only realized this because of a user report, couldn't "download refund file" because button is (correctly) not shown in this state. Desired behavior: backend should check invoice expiry and move to swap failed state immediately when invoice expires

@dni test if above scenario is fixed and web app immediately moves and shows "download refund file" button

michael1011:
Closes #367


kilrau:
typo in commit msg

maybeast (approved):

PR #771, opened 2025-01-07
feat: make transaction optional in reverse claim endpoint

### LINKED ISSUE #744: reverse swaps: settle invoice endpoint
allow clients to settle invoice without requesting any signature stuff.

can be done right now by just throwing away the sig

michael1011:
Closes #744

I really don't want another endpoint for this. Making the existing one work without a transaction seems reasonable. Wdyt @jackstar12?


jackstar12 (approved):
Altough I would prefer a separate endpoint, this seems reasonable aswell.

PR #774, opened 2025-01-08
feat: include max routing fee in submarine pairs when queried with ref

michael1011:
Needed for https://github.com/BoltzExchange/boltz-web-app/issues/784


maybeast (approved):
utACK

PR #791, opened 2025-01-20
feat: API endpoints to query historical pair data

michael1011:


boltzr/src/service/prometheus.rs
jackstar12, line 1: Have you considered any rust promotheus client libs?

package.json
jackstar12, line 1: All those `db` and `docker` feel a bit messy. Maybe we could look into using `https://github.com/BoltzExchange/regtest` in the backend aswell. Seperate issue tough

boltzr/src/service/prometheus.rs
michael1011: Yes, but I liked none of them and the code to fetch is so simple, I don't we need to include a library for that

maybeast (approved):
utACK

Issue #197, opened 2020-04-08
Boltz Backend should pay invoice via a specific route if chan_id is specified


peartobear:
Letting Boltz construct a route via a specified channel when the channel id is specified would allow for creation of more custom channel opening and normal submarine swaps.

armurbalda:
we'll look into this once there is a need again

kilrau:
To be clear: this is needed for swap INs to come in/rebalance a _specific_ channel

kilrau:
Idea just came up to also let users specify the "to be rebalenced" channel via routing hints in the invoice (even if the channels in the hints are public). With these hints it should be much easier and faster for us to construct the route.

TrezorHannes:
Let's add a CLI or ideally CLI + WebUI to determine the last hop before my node, so I can specify the incoming channel.
This probably requires for boltz-backend to probe the route again after a pubkey is set as last hop, because likely it's either
- a sink node with high cost to route to. Since other boltz channel peers would have exploited this route if it's a low PPM channel for them, you need to consider the lost opportunity costs
- a node where there is no direct channel b/w Boltz and the target incoming channel for the swapper. This would require Boltz-backend to check for possible routes, peer channel fees incurring and available liquidity

So to allow a profitable, fair and transparent process, it would require to add the incurring costs to the swap fee. Otherwise there is a risk of an exploit. For example, I could determine fixedfloat as incoming channel, 99% of liquidity is on their side. Boltz would need to route via them, which likely is a 3000ppm endevour.

kilrau:
Bumping prio here, as now this is becoming relevant for Boltz Pro

PR #797, opened 2025-01-25
feat: Lightning channels API

michael1011:
Precursor for #197


boltzr/src/api/lightning.rs
```diff
@@ -0,0 +1,300 @@
...
+    S: SwapInfos + Send + Sync + Clone + 'static,
+    M: SwapManager + Send + Sync + 'static,
+{
+    let cln = state
+        .manager
+        .get_currency(&currency)
+        .and_then(|currency| currency.cln);
+
+    Ok(match cln {
+        Some(mut cln) => {
```
maybeast, line 48: what happens in case of 0 amount?

boltzr/src/api/lightning.rs
```diff
@@ -0,0 +1,300 @@
...
+                    network: Network::Regtest,
+                    wallet: Arc::new(crate::wallet::Bitcoin::new(Network::Regtest)),
+                    cln: Some(cln.clone()),
+                    lnd: None,
+                    chain: None,
+                })
+            });
+        }
+
+        let amount = 21;
```
maybeast, line 205: want to add a test for 0 amount and overflow?

boltzr/src/service/lightning_info.rs
```diff
@@ -0,0 +1,264 @@
...
+                return Ok(channels);
+            }
+        }
+
+        let mut cln = self.get_cln(symbol)?;
+        let raw_channels: Vec<ListchannelsChannels> = cln.list_channels(Some(destination)).await?;
+
+        let mut channels = Vec::new();
+        for channel in raw_channels {
+            if !channel.public {
```
maybeast, line 177: want to add a test for this?

boltzr/src/service/lightning_info.rs
```diff
@@ -0,0 +1,264 @@
...
+    currencies: Currencies,
+    cache: Option<C>,
+}
+
+impl<C: Cache + Send + Sync> ClnLightningInfo<C> {
+    pub fn new(cache: Option<C>, currencies: Currencies) -> Self {
+        Self { cache, currencies }
+    }
+
+    async fn get_node_info_from_cln(&self, cln: &mut Cln, id: Vec<u8>) -> anyhow::Result<Node> {
```
jackstar12, line 107: Why have this extra function? `InfoFetchError` implements `From` for anyhow, so we could put it all in the `get_node_info` implementation. A bit confusing this way

boltzr/src/service/lightning_info.rs
michael1011: Funnily enough, we can't. I can' create a scenario where `public` is false. I just added that if just in case

boltzr/src/api/lightning.rs
michael1011: An invoice wiht 0 amount is fetched. We should not allowed that

boltzr/src/api/lightning.rs
michael1011: I am gonna forbid 0 amount and add a test for that. Testing overflow is testing serde which I really don't want to do

maybeast (approved):
utACK

PR #807, opened 2025-02-02
feat: different CLN amount switch for swap types

michael1011:


lib/swap/NodeSwitch.ts
```diff
@@ -35,9 +43,37 @@ class NodeSwitch {
...
+            cfg.clnAmountThreshold.reverse ||
+            NodeSwitch.defaultClnAmountThreshold,
+        };
+      }
+    } else {
+      this.clnAmountThreshold = {
+        [SwapType.Submarine]: NodeSwitch.defaultClnAmountThreshold,
+        [SwapType.ReverseSubmarine]: NodeSwitch.defaultClnAmountThreshold,
+      };
+    }
```
jackstar12, line 67:
```suggestion
    }
    this.clnAmountThreshold[SwapType.Submarine] ??= NodeSwitch.defaultClnAmountThreshold
    this.clnAmountThreshold[SwapType.Reverse] ??= NodeSwitch.defaultClnAmountThreshold
```
One can also remove the `||` operators in above this way.

maybeast (approved):

lib/swap/NodeSwitch.ts
michael1011: Can't because `this.clnAmountThreshold` itself is undefined before this statement

PR #808, opened 2025-02-03
feat: rescan for missed RSK claims on interval

michael1011:


jackstar12 (approved):

lib/wallet/ethereum/contracts/ContractEventHandler.ts
```diff
@@ -234,6 +262,43 @@ class ContractEventHandler extends TypedEventEmitter<Events> {
       },
     );
   };
+
+  private checkMissedClaims = async (provider: Provider) => {
+    this.logger.debug(
+      `Checking for missed claims ${this.networkDetails.name} from block ${this.lastClaimCheck}`,
+    );
+    const prevLastClaimCheck = this.lastClaimCheck;
+    this.lastClaimCheck = await provider.getBlockNumber();
```
maybeast, line 271: how about we update the claimCheck after the promises resolve? Additionally, do we want error handling/logging here?

michael1011: > how about we update the claimCheck after the promises resolve? 

That could cuase you to miss some blocks. Imagine you start fetching logs at block x. The from range is the last check and it will scan until the current block x. If, by the time the requests return, a new block is added, we are going to miss that one in the rescans

maybeast: what if the promises reject? in that case it'll update the lastClaimCheck without actually checking, no?

could we save block number to a local variable and only update `this.lastClaimCheck` **after** the promises resolve?

lib/wallet/ethereum/contracts/ContractEventHandler.ts
```diff
@@ -47,30 +53,58 @@ type Events = {
...
+    this.networkDetails = networkDetails;
 
     this.logger.verbose(
-      `Starting ${networkDetails.name} contract event subscriptions`,
+      `Starting ${this.networkDetails.name} contract event subscriptions`,
     );
 
     await this.subscribeContractEvents();
+
+    this.lastClaimCheck = await provider.getBlockNumber();
```
maybeast, line 91: could we miss claims here between downtime?

michael1011: No. Rescanning what we missed when we were offline is done on startup

PR #819, opened 2025-02-13
feat: sequential EVM signer

### LINKED ISSUE #818: RSK: remaining concurrent lockup issue

michael1011:
Closes #818


lib/wallet/ethereum/InjectedProvider.ts
```diff
@@ -285,8 +290,7 @@ class InjectedProvider implements Provider {
   public broadcastTransaction = async (
```
maybeast, line 290: how about logging here in case all providers fail to broadcast?

maybeast (approved):

PR #826, opened 2025-02-18
feat: nested extra fees

michael1011:
Allow clients to add extra fees to swaps by passing these parameters when creating swaps:
```json
{
    ...
    "extraFees": {
        "id": "gm",
        "percentage": 1
    }
}
```

Those extra fees added can be queried with API keys via `/v2/referral/stats/extra`:

```json
{
  "2025": {
    "2": {
      "gm": 13000
    }
  }
}
```

Do we want to allow this for every referral?

### DIFF
--- a/lib/rates/FeeProvider.ts
+++ b/lib/rates/FeeProvider.ts
@@ -8,6 +8,7 @@ import {
   splitPairId,
   stringify,
 } from '../Utils';
+import ApiErrors from '../api/Errors';
 import ElementsClient from '../chain/ElementsClient';
 import { etherDecimals, gweiDecimals } from '../consts/Consts';
 import {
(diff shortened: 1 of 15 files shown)


lib/rates/FeeProvider.ts
```diff
@@ -159,6 +160,14 @@ class FeeProvider {
     );
   };
 
+  public static calculateExtraFee = (
+    percentage: number,
+    amount: number,
+    rate: number,
+  ): number => {
+    return Math.ceil((percentage / 100) * amount * rate);
```
maybeast, line 168: Even though we validate the percentage on API request level we could consider extra input validation here.

maybeast (approved):

PR #847, opened 2025-03-10
refactor: optional Redis swap update cache

michael1011:
This PR adds optional Redis caching for swap status data, replacing in-memory storage with a persistent, distributed solution. It implements a cache abstraction with both Redis and Map-based backends, updates metrics collection accordingly, and maintains backward compatibility through optional configuration.


lib/db/Redis.ts
```diff
@@ -0,0 +1,78 @@
...
+class Redis {
+  private readonly client: ReturnType<typeof createClient>;
+
+  constructor(
+    private readonly logger: Logger,
+    config: RedisConfig,
+  ) {
+    this.client = createClient({
+      url: config.redisEndpoint,
+    });
```
maybeast, line 17:
```
// Always ensure there's a listener for errors in the client to prevent process crashes due to unhandled errors
client.on('error', error => {
    console.error(`Redis client error:`, error);
});
```

https://redis.io/docs/latest/develop/clients/nodejs/produsage/

lib/api/SwapInfos.ts
```diff
@@ -20,28 +21,32 @@ import EventHandler, { SwapUpdate } from '../service/EventHandler';
...
-  }
+  public cacheSize = async (): Promise<number> =>
+    await this.cachedSwapInfos.size();
 
   public has = async (id: string): Promise<boolean> =>
     (await this.get(id)) !== undefined;
 
   public get = async (id: string): Promise<SwapUpdate | undefined> => {
-    const cachedUpdate = this.cachedSwapInfos.get(id);
+    const cachedUpdate = await this.cachedSwapInfos.get(id);
```
maybeast, line 49: in the unlikely event that redis is down we might want to consider wrapping the get here in try/catch so the API call will still return a response. Setters should be handled by `enableOfflineQueue` which is enabled by default.

PR #850, opened 2025-03-11
Add websocket app-level ping/pong support

danielgranhao:
In the browser (e.g. WASM clients), it isn't possible to manage WebSocket ping/pong frames directly. 

This PR introduces app-level ping/pong messages to the WebSocket status stream so that clients can implement a consistent keep-alive mechanism that works across native and WASM builds.

### DIFF
--- a/boltzr/src/api/ws/status.rs
+++ b/boltzr/src/api/ws/status.rs
@@ -65,6 +65,8 @@ enum WsRequest {
     Subscribe(SubscribeRequest),
     #[serde(rename = "unsubscribe")]
     Unsubscribe(UnsubscribeRequest),
+    #[serde(rename = "ping")]
+    Ping,
 }
 
 #[derive(Deserialize, Serialize, Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


michael1011 (changes requested):
Concept ACK

Tests are missing

And we should add this new ping/pong message (with some context regarding why it's needed) to https://github.com/BoltzExchange/boltz-backend/blob/master/docs/api-v2.md#websocket

danielgranhao:
> Concept ACK
> 
> Tests are missing
> 
> And we should add this new ping/pong message (with some context regarding why it's needed) to https://github.com/BoltzExchange/boltz-backend/blob/master/docs/api-v2.md#websocket

Thanks. Test added in  [8c90816](https://github.com/BoltzExchange/boltz-backend/pull/850/commits/8c908168a34dcc0c46489549fe07ce0962f8526a) and docs updated in [a4af8eb](https://github.com/BoltzExchange/boltz-backend/pull/850/commits/a4af8eb9f116321e1d97e1b8c03f41131d08479c).

PR #862, opened 2025-03-26
refactor: transaction hooks

michael1011:


maybeast (changes requested):
missing grpc service method

PR #880, opened 2025-04-03
feat: batch-only submarine swaps

### LINKED ISSUE #752: Script-path batching of low value Liquid -> Lightning swaps
We are thinking about batching e.g. sub 1k sats Liquid -> Lightning swaps via Scriptpath (with its privacy trade-off) to make these  economically viable for wallet users.

Related: https://github.com/BoltzExchange/boltz-backend/issues/499, https://github.com/BoltzExchange/boltz-backend/issues/449

michael1011:
Closes #752

### DIFF
--- a/lib/consts/Types.ts
+++ b/lib/consts/Types.ts
@@ -29,12 +29,18 @@ export type Token = {
   decimals: number;
 };
 
-export type ChainSwapPairConfig = {
-  buyFee?: number;
-  sellFee?: number;
-
+export type PairLimits = {
   minSwapAmount?: number;
   maxSwapAmount?: number;
+};
+
+export type SubmarinePairConfig = PairLimits & {
+  minBatchedAmount?: number;
+};
+
+export type ChainSwapPairConfig = PairLimits & {
+  buyFee?: number;
+  sellFee?: number;
 };
 
 export type PairConfig = {
@@ -62,6 +68,7 @@ export type PairConfig = {
   maxSwapAmount: number;
 
   chainSwap?: ChainSwapPairConfig;
+  submarineSwap?: SubmarinePairConfig;
 
   isLegacy?: boolean;
 };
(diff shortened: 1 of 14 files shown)


lib/service/cooperative/DeferredClaimer.ts
```diff
@@ -266,6 +268,13 @@ class DeferredClaimer extends CoopSignerBase<{
     publicKey: Buffer;
     transactionHash: Buffer;
   }> => {
+    if (this.rateProvider.isBatchOnly(swap)) {
+      this.logger.debug(
+        `${swapTypeToPrettyString(swap.type)} Swap ${swap.id} is batch-only`,
+      );
+      throw Errors.NOT_ELIGIBLE_FOR_COOPERATIVE_CLAIM();
```
jackstar12, line 275: boltz-client currently doesn't consider a swap in `transaction.claim.pending` as successfull. 
https://github.com/BoltzExchange/boltz-client/blob/master/pkg/boltz/events.go#L58-L61

While throwing an error here wouldn't move the swap itself into an error state, since its just [logged as a warning](https://github.com/BoltzExchange/boltz-client/blob/436c30bd368ed31874f372f22741542ac9092867/internal/nursery/swap.go#L283), the swaps would end up in pending state longer. I think thats fine as long as we make sure to update clients before, because otherwise we would have to introduce temporary workaround which would move the state to `transaction.claimed` regardless in this case.

michael1011: This is going live next week. If you validate in the client against the existing minima, you should be just fine because it's not going to change

lib/consts/Types.ts
```diff
@@ -29,14 +29,20 @@ export type Token = {
...
-  buyFee?: number;
-  sellFee?: number;
-
+export type PairLimits = {
   minSwapAmount?: number;
   maxSwapAmount?: number;
 };
 
+export type SubmarinePairConfig = PairLimits & {
+  minBatchedAmount?: number;
```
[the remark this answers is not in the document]
michael1011: I'll keep the config consistent

lib/service/cooperative/DeferredClaimer.ts
jackstar12: Client doesnt validate the amount before swap creation (cause backend validates anyways), so one would be able to create one - but any limits from the gRPC are still the old ones.

maybeast (approved):

PR #892, opened 2025-04-09
[Draft] Payjoin Submarine Swap Consolidation

DanGould:
## Background

Greg Maxwell first mentioned [Transaction Cut-Through](https://bitcointalk.org/index.php?topic=281848.0) in 2013. The idea is that multiple transaction intents can be batched into a single transaction by coordinating before broadcast.
We applied this idea to [Boltz Exchange](https://boltz.exchange) Submarine Swaps at the MIT Bitcoin Expo Hackathon. 

Instead of just sending a transaction to the swap's lockup address, the customer initiates a spend to a Payjoin endpoint. The interactive request gives boltz-backend a chance to contribute an input from treasury in addition to the customer's funds. This consolidates UTXOs where otherwise Boltz would end up with many UTXOs, one from every single swap. To some extent this also breaks the [common input heuristic](https://en.bitcoin.it/wiki/Common-input-ownership_heuristic), the last privacy problem Satoshi Nakamoto left in the original Bitcoin whitepaper, since now, inputs come from both sender and receiver.

Since each submarine swap creates a new output for Boltz, they need to consolidate them all together to make big transfers. By consolidating during payment, they can save lots and lots of fees.

We demonstrate that the transaction flow stays almost identical to the original Submarine Swap flow, batching saves fees compared to a separate spend and consolidation, and pays the invoice the customer wished to pay.

## How we built it

We had to run both the boltz-backend and boltz-web-app on MacOS, which was tricky. Boltz.exchange is operated by one gentoo grey beard so he wasn't super motivated to make a developer experience that works across multiple environments.

Once we got the environment working, we had to write the Payjoin Manager to handle payjoin networking and transaction construction. We had to write the gRPC layer to have the main TypeScript process boltz backend is written in and the rust sidecar that handles payjoin. Once that was done, we could propagate our invoice to the front end. We had to wrangle the internal bitcoin-cli and lightning nodes to make invoices and wallets to run our payjoin cli in order to test payjoin with it but it worked.

## What's to do

- The environment setup should be upstreamed, perhaps in its own PR.
- Payjoin Manager should be integrated. Likely close to as-is
- Payjoin sessions need to be accurately persisted to continue in case of server restart
- Ideally, multiple payjoin sessions to one relay can be multiplexed
- Bitcoin RPC should use the existing RPC client implementation instead of the external dependency
- Proper automated tests that run infrastructure must be written


DanGould:
Full transparency this is NOT shaped up for code review. The first push serves as a bookmark for us to get to work making the integration work. We'll clean up the history and write nice messages so that this is a no-brainer to implement. Hoping also to receive feedback on the "What to do" list in case it's missing any major elements.

kilrau:
Anything we can do from our side to push this forward? @spacebear21 @DanGould

spacebear21:
Apologies for the slow progress lately, I was afk most of last week so didn't make progress but I'm currently working on using the existing RPC client implementation instead of `bitcoincore-rpc` which required adding some new methods to ChainClient and figuring out how to block on async RPC calls within the payjoin state machine. This is almost done so I'll aim to push it by EOD tomorrow.

Next I need to implement persistence for resumable payjoin sessions and determine how best to represent the 1:1 relationship between a swap and a payjoin session in DB. Hopefully should be able to get a working draft by end of the week.

Not directly related to this change, but it would help us in terms of velocity to make the local dev setup more macOS-friendly :) I'm currently able to test my changes locally thanks to [this hacky commit](https://github.com/BoltzExchange/boltz-backend/pull/892/commits/fac5b1cd3f4a31fb0644c301be9da4b019697a08) but like Dan said in OP we may want to review and upstream these changes into a separate PR to facilitate env setup for other devs.

maybeast:
> Not directly related to this change, but it would help us in terms of velocity to make the local dev setup more macOS-friendly :) I'm currently able to test my changes locally thanks to [this hacky commit](https://github.com/BoltzExchange/boltz-backend/pull/892/commits/fac5b1cd3f4a31fb0644c301be9da4b019697a08) but like Dan said in OP we may want to review and upstream these changes into a separate PR to facilitate env setup for other devs.

We're open to making the dev setup more macOS friendly. Have you encountered any other issues on macOS aside from the ones fixed in the hacky commit?

spacebear21:
Aside from the workarounds in that commit it's been relatively smooth. The hackiest workaround was to just remove solidity stuff from the `docker:start` command since it was in our way and not needed for the hackathon project, so there may be more issues specific to solidity/eth that we didn't run into. Happy to help test stuff on my machine if needed.

spacebear21:
Update: we ran into some difficulties implementing session persistence (full write-up [here](https://github.com/orgs/payjoin/discussions/689)), and we're making a concerted effort to address these in the core library ([draft PR](https://github.com/payjoin/rust-payjoin/pull/675/files)). Getting persistence right is the primary focus for the next PDK release so we'll keep using this implementation to inform and validate the Payjoin Session Event Log design, but will leave it in Draft until payjoin 0.24 is out.

kilrau:
> 2\. determining whether or not the hacks from [DO NOT MERGE: hacks for MacOS local dev](https://github.com/BoltzExchange/boltz-backend/pull/892/commits/a89395b470e69273e6bb565157009a8b2ea4e2ce)
>     [a89395b](https://github.com/BoltzExchange/boltz-backend/pull/892/commits/a89395b470e69273e6bb565157009a8b2ea4e2ce) are still relevant and if not, dropping them

Could you check that? @maybeast

boltzr/src/payjoin/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+}
+
+/// Try to contribute a boltz backend input to the payjoin
+async fn try_contributing_inputs(
+    payjoin: WantsInputs,
+    client: Arc<Box<dyn Client + Send + Sync>>,
+) -> Result<ProvisionalProposal, ImplementationError> {
+    let candidate_inputs = client
+        .list_unspent()
+        .await?
```
bc1cindy, line 287: these candidates come from `list_unspent()`, which calls `request("listunspent", None)` ([`chain_client.rs:209-210`](https://github.com/BoltzExchange/boltz-backend/blob/f3c117a166580e95744f0331d25af8aab0006bbd/boltzr/src/chain/chain_client.rs#L209-L210)) and `None` params serialize to an empty arg list ([`rpc_client.rs:62`](https://github.com/BoltzExchange/boltz-backend/blob/f3c117a166580e95744f0331d25af8aab0006bbd/boltzr/src/chain/rpc_client.rs#L62)), so no `minconf` is sent.

the wallet ends up confirmed-only only because it inherits Core's `minconf=1` default, and since [`ListUnspent`](https://github.com/BoltzExchange/boltz-backend/blob/f3c117a166580e95744f0331d25af8aab0006bbd/boltzr/src/chain/types.rs#L73-L81) doesn't deserialize a confirmations field there's no in-code guard. a future `minconf=0` (or a default change) would silently let the payjoin contribute 0-conf inputs, which are RBF-able/evictable and a temporal fingerprint. Worth sending `minconf` explicitly (or otherwise enforcing confirmed-only) so it doesn't depend on the RPC default?

ref: https://github.com/payjoin/rust-payjoin/issues/1597#issuecomment-4649767163

PR #893, opened 2025-04-10
feat: Add email notifications for successful + failed swaps

kaloudis:


lib/Config.ts
```diff
@@ -14,6 +14,18 @@ import type { ClnConfig } from './lightning/cln/Types';
 import type { SidecarConfig } from './sidecar/Sidecar';
 import type { NodeSwitchConfig } from './swap/NodeSwitch';
 
+type EmailConfig = {
+  enabled: boolean;
+  host?: string;
+  port?: number;
+  secure?: boolean; // true for 465, false for other ports
```
michael1011, line 21:
```suggestion
  secure?: boolean;
```

That's just for your specific provider

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,341 @@
...
+    private readonly config: EmailConfig | undefined,
+    private readonly logger: Logger,
+    private readonly service: Service,
+    private readonly walletManager: WalletManager,
+    private readonly sidecar: Sidecar,
+  ) {
+    if (!config || !config.enabled) {
+      this.logger.debug('Email notifications disabled in config');
+      return;
+    }
```
michael1011, line 41: You won't even get here when it's not enabled

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,341 @@
...
+    if (!config.host || !config.port || !config.to || !config.from) {
+      this.logger.warn(
+        'Email notifications enabled but missing required configuration (host, port, to, from)',
+      );
+      return;
+    }
+
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      this.nodemailer = require('nodemailer');
```
michael1011, line 52: We don't use require imports

lib/Config.ts
```diff
@@ -14,6 +14,18 @@ import type { ClnConfig } from './lightning/cln/Types';
 import type { SidecarConfig } from './sidecar/Sidecar';
 import type { NodeSwitchConfig } from './swap/NodeSwitch';
 
+type EmailConfig = {
+  enabled: boolean;
+  host?: string;
+  port?: number;
```
michael1011, line 20: In the `EmailNotifier` those two are *not* optional

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,341 @@
...
+      this.transporter = this.nodemailer!.createTransport({
+        host: config.host,
+        port: config.port,
+        secure:
+          config.secure === undefined ? config.port === 465 : config.secure, // Default secure based on port
```
michael1011, line 57:
```suggestion
          config.secure === undefined ? config.port === 465 : config.secure,
```

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,341 @@
...
+import type { ChainSwapInfo } from '../db/repositories/ChainSwapRepository';
+import { msatToSat } from '../lightning/ChannelUtils';
+import type Service from '../service/Service';
+import type Sidecar from '../sidecar/Sidecar';
+import type WalletManager from '../wallet/WalletManager';
+
+class EmailNotifier {
+  private transporter?: Transporter;
+  private nodemailer?: typeof import('nodemailer');
+  private enabled = false;
```
michael1011, line 26: This is redundant. `transporter` not being undefined gives you the same infornation

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,341 @@
...
+import type Swap from '../db/models/Swap';
+import type { ChainSwapInfo } from '../db/repositories/ChainSwapRepository';
+import { msatToSat } from '../lightning/ChannelUtils';
+import type Service from '../service/Service';
+import type Sidecar from '../sidecar/Sidecar';
+import type WalletManager from '../wallet/WalletManager';
+
+class EmailNotifier {
+  private transporter?: Transporter;
+  private nodemailer?: typeof import('nodemailer');
```
michael1011, line 25: This should not be a properly. It's only used in the constructor

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,341 @@
...
+        )?.networkDetails.symbol || symbol // Fallback to original symbol
+      );
+    } else {
+      return symbol;
+    }
+  };
+
+  private sendSuccessEmail = async (
+    swapDetails: AnySwap,
+    channelCreation?: any, // Use 'any' for now, refine if ChannelCreation model is available
```
michael1011, line 214: The model is available

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,341 @@
...
+      this.logger.info(
+        `Success email sent for swap ${swapDetails.id}: ${info.messageId}`,
+      );
+    } catch (error) {
+      this.logger.error(
+        `Error sending success email for swap ${swapDetails.id}: ${formatError(
+          error,
+        )}`,
+      );
+      // Re-throw or handle as needed, maybe disable temporarily?
```
michael1011, line 279: Good question

michael1011 (changes requested):
We don't use yarn

lib/Config.ts
```diff
@@ -293,6 +307,13 @@ class Config {
         interval: 1,
       },
 
+      email: {
+        enabled: false,
+        host: '',
+        to: '',
+        from: '',
+      },
+
```
michael1011, line 316: Let's omit it entirely in the default config

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,336 @@
+import nodemailer from 'nodemailer';
+import type { Transporter } from 'nodemailer';
```
michael1011, line 2: This ***has*** to be an optional dependency: https://www.webdevtutor.net/blog/typescript-optional-dependency

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,336 @@
...
+    private readonly config: EmailConfig | undefined,
+    private readonly logger: Logger,
+    private readonly service: Service,
+    private readonly walletManager: WalletManager,
+    private readonly sidecar: Sidecar,
+  ) {
+    if (!config) {
+      this.logger.debug('Email config is missing');
+      return;
+    }
```
michael1011, line 41:
```suggestion
    private readonly config: EmailConfig,
    private readonly logger: Logger,
    private readonly service: Service,
    private readonly walletManager: WalletManager,
    private readonly sidecar: Sidecar,
  ) {
```

This code will never run because the `EmailNotifier` is not initialized when `config.enabled` is not true

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,336 @@
...
+
+    if (!config.host || !config.to || !config.from) {
+      this.logger.warn(
+        'Email notifications enabled but missing required configuration (host, port, to, from)',
```
michael1011, line 45: Is the port required or not?

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,336 @@
...
+      to: this.config.to,
+      subject: subject,
+      text: `${body}${EmailNotifier.trailingWhitespace}`,
+    };
+
+    try {
+      const info = await this.transporter.sendMail(mailOptions);
+      this.logger.info(
+        `Success email sent for swap ${swapDetails.id}: ${info.messageId}`,
+      );
```
michael1011, line 269: Imho redundant but alright

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,336 @@
...
+    } catch (error) {
+      this.logger.error(
+        `Error sending success email for swap ${swapDetails.id}: ${formatError(
+          error,
+        )}`,
+      );
+    }
```
michael1011, line 276:
```suggestion
      const info = await this.transporter.sendMail(mailOptions);
      this.logger.info(
        `Success email sent for swap ${swapDetails.id}: ${info.messageId}`,
      );
```

Errors are already handled in the function that calls this one

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,336 @@
...
+        `Failure email sent for swap ${swapDetails.id}: ${info.messageId}`,
+      );
+    } catch (error) {
+      this.logger.error(
+        `Error sending failure email for swap ${swapDetails.id}: ${formatError(
+          error,
+        )}`,
+      );
+      throw error;
+    }
```
michael1011, line 332: Same thing here. Info log is imho redundant and the function calling this one handles the error

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,336 @@
...
+
+    const mailOptions = {
+      from: this.config.from,
+      to: this.config.to,
+      subject: subject,
+      text: `${body}${EmailNotifier.trailingWhitespace}`,
+    };
+
+    try {
+      const info = await this.transporter.sendMail(mailOptions);
```
michael1011, line 321: This is redundant. Let's do a `sendEmail` method

michael1011 (changes requested):
Should be good to merge after

lib/notifications/EmailNotifier.ts
```diff
@@ -0,0 +1,312 @@
...
+import type Swap from '../db/models/Swap';
+import type { ChainSwapInfo } from '../db/repositories/ChainSwapRepository';
+import { msatToSat } from '../lightning/ChannelUtils';
+import type Service from '../service/Service';
+import type Sidecar from '../sidecar/Sidecar';
+import type WalletManager from '../wallet/WalletManager';
+
+class EmailNotifier {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+  private transporter?: import('nodemailer').Transporter;
```
kaloudis, line 25: only question I have left is if we can find a better way to do this

michael1011:
```diff
diff --git a/lib/notifications/EmailNotifier.ts b/lib/notifications/EmailNotifier.ts
index 86b84fae..be2adb4f 100644
--- a/lib/notifications/EmailNotifier.ts
+++ b/lib/notifications/EmailNotifier.ts
@@ -1,3 +1,4 @@
+import type { Transporter } from 'nodemailer';
 import type { EmailConfig } from '../Config';
 import { satoshisToSatcomma } from '../DenominationConverter';
 import type Logger from '../Logger';
@@ -21,8 +22,7 @@ import type Sidecar from '../sidecar/Sidecar';
 import type WalletManager from '../wallet/WalletManager';

 class EmailNotifier {
-  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
-  private transporter?: import('nodemailer').Transporter;
+  private transporter?: Transporter;

   // This is a hack to add trailing whitespace which is trimmed by default in some email clients
   private static trailingWhitespace = '\n\n';
   ```

michael1011 (approved):
Good to be merged after the type import is fixed

michael1011 (approved):

PR #913, opened 2025-04-25
feat: add swap id to transaction hook

maybeast:


proto/boltzrpc.proto
```diff
@@ -386,18 +386,19 @@ message SwapCreation {
...
 message TransactionHookResponse {
-  string id = 1;
+  string id = 1; // swap id
   Action action = 2;
 }
 
 message TransactionHookRequest {
   string symbol = 1;
-  string id = 2;
+  string id = 2; // swap id
```
michael1011, line 395:
```suggestion
  // Swap ID
  string id = 1;
```

lib/swap/hooks/TransactionHook.ts
```diff
@@ -13,21 +13,23 @@ class TransactionHook extends Hook<
 
   public hook = (
     symbol: string,
-    id: string,
+    txId: string,
     tx: Buffer,
     confirmed: boolean,
+    swapId: string,
```
michael1011, line 19: Let's make this the first parameter

lib/swap/UtxoNursery.ts
```diff
@@ -283,6 +284,7 @@ class UtxoNursery extends TypedEventEmitter<{
             SwapUpdateEvent.InvoiceSet,
             SwapUpdateEvent.TransactionMempool,
             SwapUpdateEvent.TransactionZeroConfRejected,
+            SwapUpdateEvent.TransactionConfirmed,
```
michael1011, line 287: Do we have to do that for chain swaps too?

michael1011 (approved):

PR #933, opened 2025-05-15
fix: bitcoind/elementsd cookie mounting on MacOS

maybeast:
Docker desktop doesn't properly mount the bitcoind/elementsd cookies on MacOS. This PR implements a potential workaround.


kilrau:
Wanna test? @spacebear21 @DanGould

michael1011 (changes requested):
Let's stop mounting cookies completely and use this instead: https://github.com/bitcoin/bitcoin/blob/master/share/rpcauth/rpcauth.py

PR #959, opened 2025-06-03
Add extraFee percentage to percentageFee before calculation

danielgranhao:
This PR proposes changing the way extra fees are calculated so that clients don't need to be aware of them. 

By calculating the total fees by adding the relative fees first (`percentageFee` and `extraFee`), clients can be informed of a single relative fee (the sum) and still get to the same value as the server. As it was, clients would need to know the extra fee in order to calculate `ceil(percentageFee * amount) + ceil(extraFee * amount)`.

### DIFF
--- a/lib/rates/FeeProvider.ts
+++ b/lib/rates/FeeProvider.ts
@@ -8,7 +8,6 @@ import {
   splitPairId,
   stringify,
 } from '../Utils';
-import ApiErrors from '../api/Errors';
 import ElementsClient from '../chain/ElementsClient';
 import { etherDecimals, gweiDecimals } from '../consts/Consts';
 import {
@@ -147,18 +146,6 @@ class FeeProvider {
     );
   };
 
-  public static calculateExtraFee = (
-    percentage: number,
-    amount: number,
-    rate: number,
-  ): number => {
-    if (percentage < 0) {
-      throw ApiErrors.INVALID_EXTRA_FEES_PERCENTAGE(percentage);
-    }
-
-    return Math.ceil((percentage / 100) * amount * rate);
-  };
-
   public init = (pairs: PairConfig[]): void => {
     pairs.forEach((pair) => {
       const pairId = getPairId(pair);
@@ -251,7 +238,7 @@ class FeeProvider {
     percentageFee: number;
     extraFee?: number;
   } => {
-    let percentageFee = this.getPercentageFee(
+    const percentageFeeCalculation = this.getPercentageFee(
       pair,
       orderSide,
       type,
@@ -259,8 +246,10 @@ class FeeProvider {
       referral,
     );
 
-    if (percentageFee !== 0) {
-      percentageFee = percentageFee * amount * rate;
+    let percentageFee = 0;
+
+    if (percentageFeeCalculation !== 0) {
+      percentageFee = Math.ceil(percentageFeeCalculation * amount * rate);
     }
 
     const { base, quote } = splitPairId(pair);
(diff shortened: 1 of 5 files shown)


lib/rates/FeeProvider.ts
```diff
@@ -274,16 +263,23 @@ class FeeProvider {
...
     if (extraFees !== undefined) {
-      extraFee = FeeProvider.calculateExtraFee(
-        extraFees.percentage,
-        amount,
-        rate,
-      );
+      // Clients may only be informed of a single percentage fee resulting from the addition of both percentage fees.
+      // The addition is done with the percentage values to avoid discrepancies due to floating point precision issues.
+      let percentageFeeWithExtraFees =
+        (percentageFeeCalculation * 100 + extraFees.percentage) / 100;
```
michael1011, line 269: Does that actually fix the floating point issue? I think we will only ever use 2 decimals of the %, so `* 10_000` and working with a BigInt would be better imho

lib/rates/FeeProvider.ts
```diff
@@ -274,16 +263,23 @@ class FeeProvider {
...
+
+      if (percentageFeeWithExtraFees !== 0) {
+        percentageFeeWithExtraFees = Math.ceil(
+          percentageFeeWithExtraFees * amount * rate,
+        );
+      }
+
+      extraFee = percentageFeeWithExtraFees - percentageFee;
```
michael1011, line 277: This is JavaScript, this value needs to be rounded

lib/service/Renegotiator.ts
```diff
@@ -260,14 +259,20 @@ class Renegotiator {
...
     if (extraFees !== undefined && extraFees !== null) {
-      extraFee = FeeProvider.calculateExtraFee(
-        extraFees.percentage,
-        swap.receivingData.amount!,
+      const serverLockAmountWithExtraFees = this.calculateServerLockAmount(
         pair.rate,
+        swap.receivingData.amount!,
+        // Clients may only be informed of a single percentage fee resulting from the addition of both percentage fees.
+        // The addition is done with the percentage values to avoid discrepancies due to floating point precision issues.
+        (feePercent * 100 + extraFees.percentage) / 100,
```
michael1011, line 267: Same there regarding the * 100 / 100 trickery

lib/service/Renegotiator.ts
```diff
@@ -260,14 +259,20 @@ class Renegotiator {
...
+        baseFee,
       );
-      serverLockAmount.serverLockAmount = Math.floor(
-        serverLockAmount.serverLockAmount - extraFee,
-      );
+
+      extraFee =
+        serverLockAmount.serverLockAmount -
+        serverLockAmountWithExtraFees.serverLockAmount;
```
michael1011, line 273: Needs to be rounded too

lib/service/Service.ts
```diff
@@ -1842,57 +1897,38 @@ class Service {
...
-        holdInvoiceAmount,
+      const {
+        onchainAmount: onchainAmountWithExtraFees,
+        holdInvoiceAmount: holdInvoiceAmountWithExtraFees,
+      } = this.calculateReverseSwapAmounts(
+        args,
         rate,
+        // Clients may only be informed of a single percentage fee resulting from the addition of both percentage fees.
+        // The addition is done with the percentage values to avoid discrepancies due to floating point precision issues.
+        (feePercent * 100 + args.extraFees.percentage) / 100,
```
michael1011, line 1920: Same here

lib/service/Service.ts
```diff
@@ -1842,57 +1897,38 @@ class Service {
...
+        baseFee,
       );
 
       if (invoiceAmountDefined) {
-        onchainAmount = Math.floor(onchainAmount - extraFee);
+        extraFee = onchainAmount - onchainAmountWithExtraFees;
       } else {
-        holdInvoiceAmount = Math.ceil(holdInvoiceAmount + extraFee);
+        extraFee = holdInvoiceAmountWithExtraFees - holdInvoiceAmount;
```
michael1011, line 1927: Rounding

lib/service/Service.ts
```diff
@@ -2157,64 +2265,56 @@ class Service {
...
+          serverLockAmount: serverLockAmountWithExtraFees,
+        } = this.calculateChainSwapAmounts(
+          {
+            userLockAmount: originalUserLockAmount,
+            serverLockAmount: originalServerLockAmount,
+          },
           rate,
+          // Clients may only be informed of a single percentage fee resulting from the addition of both percentage fees.
+          // The addition is done with the percentage values to avoid discrepancies due to floating point precision issues.
+          (feePercent * 100 + args.extraFees.percentage) / 100,
```
michael1011, line 2306: Same

lib/service/Service.ts
```diff
@@ -2157,64 +2265,56 @@ class Service {
...
+          baseFee,
         );
 
         if (userLockAmountDefined) {
-          args.serverLockAmount = Math.floor(args.serverLockAmount - extraFee);
+          extraFee = args.serverLockAmount - serverLockAmountWithExtraFees;
         } else {
-          args.userLockAmount = Math.ceil(args.userLockAmount + extraFee);
+          extraFee = userLockAmountWithExtraFees - args.userLockAmount;
```
michael1011, line 2313: Rounding

michael1011 (changes requested):

PR #986, opened 2025-07-01
refactor: improve MRH watcher performance

michael1011:
This commit improves the performance of the MRH watcher by only checking unconfirmed transactions and avoids lagged channels by increasing the capacity.

## Summary by CodeRabbit

* **New Features**
  * Transactions now include a confirmation status flag, allowing users to distinguish between confirmed and unconfirmed transactions.
  * Added support for retrieving detailed transaction information, including confirmation count.

* **Improvements**
  * Enhanced transaction monitoring to ignore confirmed transactions where appropriate.
  * Increased transaction channel size for improved performance and reliability.
  * Added cancellation support to client creation for better control over background tasks.

* **Bug Fixes**
  * Improved cancellation support for client operations, providing more responsive shutdowns.

* **Tests**
  * Updated and added tests to cover new transaction details and confirmation status.
  * Converted relevant tests to run asynchronously for better reliability and coverage.

### DIFF
--- a/boltzr/src/chain/zmq_client.rs
+++ b/boltzr/src/chain/zmq_client.rs
@@ -7,6 +7,8 @@ use tokio::sync::broadcast::{self, Sender};
 use tracing::{debug, error, warn};
 use zeromq::{Socket, SocketRecv, ZmqError, ZmqMessage};
 
+pub const ZMQ_TX_CHANNEL_SIZE: usize = 1024 * 16;
+
 #[derive(Debug, Clone)]
 pub struct ZmqClient {
     client_type: Type,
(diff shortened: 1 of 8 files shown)


boltzr/src/chain/zmq_client.rs
```diff
@@ -7,6 +7,8 @@ use tokio::sync::broadcast::{self, Sender};
 use tracing::{debug, error, warn};
 use zeromq::{Socket, SocketRecv, ZmqError, ZmqMessage};
 
+pub const ZMQ_TX_CHANNEL_SIZE: usize = 1024 * 16;
```
jackstar12, line 10: isnt this a bit much considering the fact that we have another layer for the transaction confirmations now?

jackstar12 (approved):

boltzr/src/chain/zmq_client.rs
michael1011: Better safe than lagged

PR #987, opened 2025-07-01
fix: reset pending sweeps gauges to zero

michael1011:
## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured that all currency symbols have a default gauge value initialized in metrics before reporting pending sweep counts and amounts. This prevents missing or uninitialized values for any currency.


michael1011:
@coderabbitai review

PR #988, opened 2025-07-01
fix: encoding of MRH in swapinfo responses

michael1011:


jackstar12 (approved):
A local var for the routingHint would be nice - very verbose right now

PR #993, opened 2025-07-07
feat: run MRH watcher through 0-conf API

michael1011:


boltzr/src/chain/elements_client.rs
```diff
@@ -171,4 +275,58 @@ pub mod test {
...
+
+        let received = zero_conf_safe.await.unwrap();
+        assert!(!received);
+
+        generate_block(&client).await;
+    }
+
+    #[tokio::test]
+    #[serial(LBTC)]
+    async fn test_zero_conf_safe_tool() {
```
jackstar12, line 311: lets add a case where its considered safe here

PR #1006, opened 2025-07-14
fix: fail self payments when reverse swap is refunded

michael1011:
## Summary by CodeRabbit

* **New Features**
  * Added a new property to refund transactions that indicates when a refund is finalized.

* **Bug Fixes**
  * Improved error handling for self payments involving refunded transactions.

* **Tests**
  * Introduced integration tests for refund transaction finality.
  * Added unit tests for self payment logic when handling refunded transactions.

### DIFF
--- a/lib/lightning/SelfPaymentClient.ts
+++ b/lib/lightning/SelfPaymentClient.ts
@@ -17,6 +17,7 @@ import type LightningPayment from '../db/models/LightningPayment';
 import type ReverseSwap from '../db/models/ReverseSwap';
 import { NodeType } from '../db/models/ReverseSwap';
 import type Swap from '../db/models/Swap';
+import RefundTransactionRepository from '../db/repositories/RefundTransactionRepository';
 import ReverseSwapRepository from '../db/repositories/ReverseSwapRepository';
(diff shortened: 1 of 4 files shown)


lib/lightning/SelfPaymentClient.ts
```diff
@@ -278,6 +279,16 @@ class SelfPaymentClient
         feeMsat: 0,
         preimage: getHexBuffer(reverseSwap.preimage),
       };
+    } else if (reverseSwap.status === SwapUpdateEvent.TransactionRefunded) {
+      const refundTx = await RefundTransactionRepository.getTransactionForSwap(
+        reverseSwap.id,
+      );
+      if (refundTx !== null && refundTx !== undefined && refundTx.isFinal) {
```
jackstar12, line 286: why only throw when the refund tx is final?

michael1011: So that we only allow cooperatively refunding the lockup of the submarine swap after we are sure the refund of the reverse swap is final and can't be replaced anymore

jackstar12 (approved):

maybeast (approved):

PR #1027, opened 2025-07-31
fix: swap rescue key drift

michael1011:
Chain swaps which have two keys (claim and rescue) could error when one of their keys was in a lookup batch but the other was not (eg, batch size 50; claim key is index 49 but the refund one index 50)

## Summary by CodeRabbit

* **New Features**
  * Added a new API endpoint to restore swaps created with an XPUB, enabling swap history recovery.
  * Introduced detailed schemas for restorations, including claim and refund details.
* **Improvements**
  * Enhanced swap handling with uniform ID access and deduplication.
  * Extended key derivation range for improved key coverage.
  * Updated API documentation with clearer descriptions and schema references.
* **Bug Fixes**
  * Simplified and corrected swap filtering and key lookup logic.
* **Tests**
  * Updated tests to align with new key derivation behavior and method signatures.

### DIFF
--- a/boltzr/src/service/rescue.rs
+++ b/boltzr/src/service/rescue.rs
@@ -21,6 +21,10 @@ use tracing::{debug, instrument, trace};
 const DERIVATION_PATH: &str = "m/44/0/0/0";
 const GAP_LIMIT: u64 = 50;
 
+trait Identifiable {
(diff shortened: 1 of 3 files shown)


kiocos (approved):
tACK

PR #1032, opened 2025-08-02
refactor: RBF bump refund transactions in Rust

michael1011:
Still WIP but really happy we can construct transactions in the sidecar now


maybeast (approved):

boltzr/src/config.rs
```diff
@@ -7,6 +7,8 @@ use tracing::{debug, info, trace};
 #[derive(Deserialize, Serialize, PartialEq, Clone, Debug)]
 pub struct CurrencyConfig {
     pub symbol: String,
+    #[serde(rename = "preferredWallet")]
+    pub preferred_wallet: Option<String>,
```
jackstar12, line 11: why already add that config value if we only support core right now?

boltzr/src/db/models/chain_swap.rs
```diff
@@ -120,6 +127,117 @@ impl SomeSwap for ChainSwapInfo {
     fn status(&self) -> SwapUpdate {
         SwapUpdate::parse(self.swap.status.as_str())
     }
+
+    fn refund_symbol(&self) -> Result<String> {
+        Ok(self.sending().symbol.clone())
+    }
+
+    async fn refund_details(
```
jackstar12, line 135: i think this function does way too much, considering that its part of a db model. we should try to get more of this logic into boltz-core since its got devent overlap with the reverse swap equivalent and is kinda hard to read right now.

boltz-core/src/elements/tx.rs
```diff
@@ -580,6 +580,15 @@ mod tests {
 
     const FUNDING_AMOUNT: u64 = 100_000;
 
+    // There is some variance in the proof sizes of Elements transactions
+    fn assert_within_one_percent(actual: u64, expected: u64) {
+        let tolerance = expected / 100;
+        let lower_bound = expected.saturating_sub(tolerance);
+        let upper_bound = expected.saturating_add(tolerance);
+
+        assert!(actual >= lower_bound && actual <= upper_bound);
```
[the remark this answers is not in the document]
michael1011: Yes, it can be that large because we intentially overshoot

PR #1033, opened 2025-08-04
feat: Ark rescan after stream disconnect

michael1011:


lib/chain/ArkSubscription.ts
```diff
@@ -0,0 +1,232 @@
...
+
+        for (const vhtlc of res.vhtlcsList) {
+          this.emit('vhtlc.created', {
+            address,
+            txId: vhtlc.outpoint!.txid,
+            vout: vhtlc.outpoint!.vout,
+            amount: vhtlc.amount,
+          });
+
+          if (vhtlc.spentBy !== '') {
```
louisinger, line 107: nit: you may prefer to check `is_spent` boolean ?

lib/chain/ArkSubscription.ts
```diff
@@ -0,0 +1,232 @@
...
+      this.vHtlcStream.cancel();
+      this.vHtlcStream = undefined;
+    }
+
+    while (true) {
+      await sleep(ArkSubscription.reconnectInterval);
+
+      try {
+        const status = await this.client.getWalletStatus();
+        if (!status.synced) {
```
louisinger, line 208: it should also check if fulmine is unlocked (`status.unlocked`)

PR #1038, opened 2025-08-15
fix: Bump ZMQ Inactivity Timeout

SwapMarket:
Increase the timeout to 5 minutes. One minute was too short for Elements and caused false positives.

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased the inactivity timeout for ZMQ subscriptions on non-Regtest networks to reduce unintended timeouts during low activity, improving stability of real-time updates. Regtest behavior is unchanged.


michael1011 (changes requested):
Also needs to be bumped here: https://github.com/BoltzExchange/boltz-backend/blob/master/lib/chain/ZmqClient.ts#L43

And how about going for something like 5 minutes instead? To completely avoid false positivies in the future?

michael1011 (approved):

PR #1045, opened 2025-08-21
docs: link to boltz-client for go examples

jackstar12:
raw code snippets are very easy to get outdated


kilrau (changes requested):
I wouldn't want to do this, it's lazy. Did you check options to bring in the external files at build time? Quick GPT search brought up 3 ways to do that:

Image

Imo we don't need the scheduled build, we update and build API docs regularly so as long as it gets picked up then automatically it's enough.

If you think that's absolutely not viable, well I'd at lesat keep the `Go Bitcoin` tab and simply link to each respective example like https://github.com/BoltzExchange/boltz-client/blob/master/examples/submarine/submarine.go there.

Also you broke the submarine swap tabs somehow, adding empty duplicates:
Image

docs/api-v2.md
```diff
@@ -628,156 +634,6 @@ const submarineSwap = async () => {
...
-
-func main() {
-	if err := submarineSwap(); err != nil {
-		fmt.Println(err)
-		os.Exit(1)
-	}
-}
-```
-
-:::
```
kilrau, line 779: that was the issue

michael1011:
Superseded by #1047

PR #1064, opened 2025-09-10
docs: add pro stats endpoint to swagger

kilrau:
Went for a separate section for Pro as I didn't find `Info` nor `Nodes` nor `Referral` fitting and it's the only endpoint that requires setting the pro referral.


michael1011 (changes requested):
*Never* edit the swagger file directly. It is generated with `npm run swagger` based on the Router TypeScript files

PR #1076, opened 2025-09-21
feat: `cltvDelta` config option

jackstar12:


lib/service/TimeoutDeltaProvider.ts
```diff
@@ -70,6 +70,7 @@ class TimeoutDeltaProvider {
     private readonly sidecar: Sidecar,
     private readonly currencies: Map<string, Currency>,
     private readonly nodeSwitch: NodeSwitch,
+    private readonly swapConfig: SwapConfig,
```
michael1011, line 73: Pick the `cltvDelta` property you need: https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

michael1011 (changes requested):
Tests are broken

PR #1127, opened 2025-11-20
Swap Restore follow-ups

### LINKED ISSUE #1123: Restore with public key batch instead of xpub
Builds on top of #1107 

Which allows swap restore with:
- xpub
- and singular public keys

Restoring via a batch of public keys should also be possible to avoid the client giving us their xpub while still having reasonable performance

michael1011:
Closes #1123

## Summary by CodeRabbit

* **New Features**
  * Rescue/restore endpoints accept an xpub, a single public key, or multiple public keys (oneOf input).
  * Added POST /swap/restore/index to compute the highest derivation index for provided keys.
  * Configurable gapLimit for tuning key scanning; input validation enforces allowed range.
  * Recoverable swap responses now include optional preimageHash.

* **Other**
  * OpenAPI/docs updated to reflect new request shapes, response schema, and validation.

### DIFF
--- a/boltzr/src/service/pubkey_iterator.rs
+++ b/boltzr/src/service/pubkey_iterator.rs
@@ -123,7 +123,7 @@ impl PubkeyIterator for KeyVecIterator {
     }
 
     fn gap_limit(&self) -> u32 {
-        DEFAULT_GAP_LIMIT
+        self.max_keys()
     }
 
     fn derive_keys(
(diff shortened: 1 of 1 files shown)


kiocos (commented):
tACK for custom gap limit, derivation path and the new `/restore/index` endpoint.
Haven't tested using public key(s) for derivation.

boltzr/src/api/rescue.rs
```diff
@@ -1,59 +1,121 @@
...
+        gap_limit: Option<u32>,
+    },
+    PublicKey {
+        #[serde(rename = "publicKey")]
+        public_key: PublicKeyDeserialize,
+    },
+    PublicKeyVec {
+        #[serde(rename = "publicKeys")]
+        public_keys: PublicKeyVecDeserialize,
+    },
```
jackstar12: We really need a dedicated variant for a single key? `PublicKeys` would be just fine imo

PR #1150, opened 2025-12-02
feat: sync key index incrementation at database level

jackstar12:
## Summary by CodeRabbit

* **Refactor**
  * Wallet key provisioning is now asynchronous and uses a simplified initialization signature.
  * Key index management moved to an atomic increment-and-return operation that yields the current index or undefined.
  * Swap flows updated to await key retrievals for consistent key/index usage.

* **Tests**
  * Unit and integration tests updated to await async key retrieval and to match the simplified init behavior.

### DIFF
--- a/test/integration/Core.spec.ts
+++ b/test/integration/Core.spec.ts
@@ -47,6 +47,7 @@ import { CurrencyType, SwapType, SwapVersion } from '../../lib/consts/Enums';
 import Database from '../../lib/db/Database';
 import type ChainSwapData from '../../lib/db/models/ChainSwapData';
 import type Swap from '../../lib/db/models/Swap';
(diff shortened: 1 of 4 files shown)


lib/db/repositories/KeyRepository.ts
```diff
@@ -22,20 +23,22 @@ class KeyRepository {
...
+        bind: [symbol],
+        type: QueryTypes.SELECT, // has to be select for returning the new index
       },
     );
+
+    if (!results || results.length === 0) {
+      return undefined;
+    }
+
+    return (results[0] as { highestUsedIndex: number }).highestUsedIndex;
```
michael1011:
```suggestion
    return results[0].highestUsedIndex;
```

PR #1174, opened 2025-12-30
feat: sponsor refunds for refunds of liquid assets

### LINKED ISSUE #1091: Endpoint allowing to rescue non-LBTC assets
Client constructs a sighash single psbt and gives it to us. we add an input to pay the fee, blind it and broadcast. For now Liquid specific, but endpoint should be generic enough to allow for other non-Liquid layers with the same issue in future too.

Not sure if adding to `/swap/restore` or a new endpoint makes more sense.

michael1011:
Closes #1091

## Summary by CodeRabbit

* **New Features**
  * Asset Rescue: service and HTTP endpoints to prepare and broadcast rescue transactions (MuSig-assisted signing, UTXO reservation).
  * Chain/wallet RPCs now support wallet-scoped operations and return unspent/signing responses.
  * Cache: added take/delete operations and background cleanup for expired entries.

* **Documentation**
  * OpenAPI/Swagger updated with Asset Rescue endpoints and schemas.

* **Other**
  * Musig API refactored to a staged builder surface; wallet address API now accepts an optional wallet parameter.

### DIFF
--- a/boltzr/src/cache/memcache.rs
+++ b/boltzr/src/cache/memcache.rs
@@ -1,12 +1,18 @@
+use crate::utils::{DropGuard, defer};
 use dashmap::DashMap;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use tokio::task::JoinHandle;
+use tokio::time;
 
-#[derive(Clone, Debug)]
+const CLEANUP_INTERVAL: Duration = Duration::from_secs(60);
+
+#[derive(Debug, Clone)]
 pub struct MemCache {
     pub map: Arc<DashMap<String, String>>,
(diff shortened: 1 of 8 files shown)


boltzr/src/db/models/mod.rs
```diff
@@ -154,6 +154,8 @@ pub trait SomeSwap {
 
     fn sending_outpoint(&self) -> Result<Option<Outpoint>>;
 
+    fn asset_rescue_details(&self) -> Result<(String, u64, Vec<u8>, ElementsTree)>;
```
jackstar12: This deserves an explicit type. In fact, why not move that logic entirely to https://github.com/BoltzExchange/boltz-backend/pull/1174/changes#diff-1543227a36a94a93cafc521009b5cb066b720fa0379b4e140bcbd04a551a6a6eR434-R452 and avoid extra stuff in the generic models.

PR #1210, opened 2026-01-20
feat: sub-sat fee estimation

### LINKED ISSUE #1108: Lower min sat/vbyte
We should allow for this after mempool shipped sub-sat estimations (https://x.com/OrangeSurfBTC/status/1984117869882560665).

kiocos:
Closes #1108 

## Summary by CodeRabbit

* **New Features**
  * Configurable per-chain minimum fee floor for fee estimation.

* **Improvements**
  * Fee estimates now respect chain-specific floors and return three-decimal precision.
  * External fee source integration updated to use more precise fee data and include an optional economy rate.
  * When external estimates are missing, the configured fee floor is used instead of a hardcoded default.

* **Tests**
  * Tests updated to allow tolerance-based fee assertions rather than strict equality.

### DIFF
--- a/boltzr/src/chain/utils.rs
+++ b/boltzr/src/chain/utils.rs
@@ -334,7 +334,15 @@ mod test {
 
         let client = Arc::new(client) as Arc<dyn crate::chain::Client + Send + Sync>;
         let fee = tx.calculate_fee(&client).await.unwrap();
-        assert_eq!(fee, tx.vsize());
+
+        let expected = tx.vsize();
+        let tolerance = expected / 10;
+        assert!(
+            fee >= expected.saturating_sub(tolerance) && fee <= expected + tolerance,
+            "fee {} not within ±10% of expected vsize {}",
+            fee,
+            expected
+        );
     }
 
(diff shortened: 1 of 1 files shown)


lib/chain/ChainClient.ts
```diff
@@ -276,7 +276,7 @@ class ChainClient extends BaseClient implements IChainClient {
           `'estimatesmartfee' method not found on ${this.symbol} chain`,
         );
 
-        return 2;
+        return 0.2;
```
michael1011: Can't we use the `this.feeFloor`?

lib/chain/ChainClient.ts
```diff
@@ -254,7 +254,7 @@ class ChainClient extends BaseClient implements IChainClient {
         ? mempoolFee
         : await this.estimateFeeChainClient(confTarget);
 
-    return Math.max(estimation, this.feeFloor);
+    return Math.max(Number(estimation), this.feeFloor);
```
michael1011: ... then you don't need to convert it to Number here

PR #1228, opened 2026-01-25
feat: funding address claimer

jackstar12:
## Summary by CodeRabbit

* **New Features**
  * Added support for funding address-based batch claims on Bitcoin and Liquid chains.

* **Changes**
  * Batch claim responses now include multiple transaction IDs and per-swap fees instead of single aggregated values.
  * Added virtual size calculation for transactions.
  * Updated transaction version handling.

### DIFF
--- a/boltz-core/src/bitcoin/tx.rs
+++ b/boltz-core/src/bitcoin/tx.rs
@@ -81,6 +81,8 @@ pub fn construct_raw<C: Signing + Verification>(
     };
 
     let mut tx = Transaction {
+        // Use V3 transactions to enable zero-fee presigned funding-address transactions
+        // that can be spent by a fee-paying child transaction via TRUC relay.
         version: Version(3),
         lock_time: if let Some(lock_time) = inputs
             .iter()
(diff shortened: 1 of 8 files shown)


michael1011:
@coderabbitai review

boltz-core/src/bitcoin/tx.rs
```diff
@@ -81,7 +81,7 @@ pub fn construct_raw<C: Signing + Verification>(
     };
 
     let mut tx = Transaction {
-        version: Version::TWO,
+        version: Version(3),
```
michael1011: Let's add a comment why we are using V3

boltzr/src/service/funding_address_claimer.rs
```diff
@@ -0,0 +1,522 @@
...
+                    };
+                    let (tx, _) = construct_tx(&Params::Bitcoin(params.clone()))?;
+
+                    let btc_tx = match tx {
+                        Transaction::Bitcoin(tx) => tx,
+                        Transaction::Elements(_) => {
+                            return Err(anyhow!("expected Bitcoin transaction"));
+                        }
+                    };
+                    let claim_vsize = btc_tx.vsize() as u64;
```
michael1011: This almost calls for a trait on the transaction enum to avoid the match above. All you need is the vsize

boltzr/src/swap/manager.rs
```diff
@@ -303,7 +311,25 @@ impl SwapManager for Manager {
             .ok_or_else(|| anyhow!("chain client not found"))?;
         let wallet = currency.wallet.ok_or_else(|| anyhow!("wallet not found"))?;
 
-        let inputs: Vec<InputDetail> = futures::stream::iter(swaps)
+        let claimable = self
+            .funding_address_claimer
+            .funding_claimable_swaps(submarines, chain_swaps)
+            .await?;
```
michael1011: This is really weird. Does that `claimable` mean they are claimable as funding addresses?

PR #1236, opened 2026-01-30
refactor: funding address review

jackstar12:
- **refactor: funding address review**
- **wipwip**

## Release Notes

* **API Changes**
  * Renamed funding address response field from "boltzPublicKey" to "serverPublicKey"
  * Funding address creation now returns HTTP 201 Created instead of 200 OK
  * Funding address signature endpoint now uses PATCH method instead of POST
  
* **Improvements**
  * Enhanced WebSocket subscription handling with improved reliability

### DIFF
--- a/boltzr/src/api/ws/subscriptions.rs
+++ b/boltzr/src/api/ws/subscriptions.rs
@@ -553,7 +553,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_inject_updates_sends_empty_vec_when_no_matching_swaps() {
+    async fn test_inject_updates_sends_nothing_when_no_matching_swaps() {
         let (subscriptions, _, _) = create_status_subscriptions();
         let connection_id = 1;
         let mut rx = subscriptions.connection_added(connection_id);
@@ -564,8 +564,7 @@ mod tests {
 
         subscriptions.inject_updates(connection_id, updates).await;
 
-        let received = rx.recv().await.unwrap();
-        assert_eq!(received.len(), 0);
+        assert!(rx.try_recv().is_err());
     }
 
     #[test]
--- a/boltzr/src/api/ws/utils.rs
+++ b/boltzr/src/api/ws/utils.rs
@@ -6,6 +6,10 @@ const SEND_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Sends updates through an mpsc channel with a timeout to prevent blocking.
 pub async fn send_with_timeout<T: Send + 'static>(tx: mpsc::Sender<Vec<T>>, updates: Vec<T>) {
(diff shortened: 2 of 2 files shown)


boltzr/src/api/ws/subscriptions.rs
```diff
@@ -0,0 +1,529 @@
...
+        updates
+            .iter()
+            .filter(|update| ids.contains(&update.id().to_string()))
+            .cloned()
+            .collect()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
michael1011: Can't we just keep the old tests instead of this claude slop?

michael1011 (changes requested):
What is commit message "wipwip" supposed to be?

The swagger script has not been run and the generated JSON is outdated

boltzr/src/api/ws/subscriptions.rs
```diff
@@ -0,0 +1,529 @@
...
+                    }
+                }
+            }
+        });
+    }
+
+    fn filter_updates(ids: &[String], updates: &[T]) -> Vec<T> {
+        updates
+            .iter()
+            .filter(|update| ids.contains(&update.id().to_string()))
```
michael1011: Does that need `to_string()`?

boltzr/src/api/ws/subscriptions.rs
jackstar12: Just matched the old test suite with minimal adjustments now.

test/integration/swap/UtxoNursery.spec.ts
```diff
@@ -660,15 +663,39 @@ describe('UtxoNursery', () => {
...
+      );
+      const tx = parseTransaction<LiquidTransaction>(
+        CurrencyType.Liquid,
+        await elementsClient.getRawTransaction(txId),
+      );
+
+      const vout = tx.outs.findIndex((out) => {
+        try {
+          const address = (elementsClient as any).encodeAddress(out.script);
+          return address === fundingAddressAddr;
```
michael1011: Decode the address once and then compare the script pubkey

boltzr/src/swap/status.rs
```diff
@@ -76,9 +76,10 @@ pub fn serialize_swap_updates(
     status.iter().map(|status| status.to_string())
 }
 
-#[derive(EnumString, Display, PartialEq, Clone, Copy, Debug)]
+#[derive(EnumString, Display, PartialEq, Clone, Copy, Debug, Default)]
 pub enum FundingAddressStatus {
     #[strum(serialize = "created")]
+    #[default]
```
michael1011: I would prefer no default for this to avoid having the Created status by accident

PR #1288, opened 2026-02-27
feat: funding address amount validation

jackstar12:


boltzr/src/service/funding_address_signer.rs
```diff
@@ -295,20 +306,30 @@ impl FundingAddressSigner {
...
-            .map(|swap| SwapInfo {
-                status: swap.status(),
-                lockup_address: swap.lockupAddress,
-                timeout_block_height: swap.timeoutBlockHeight as u32,
+            .and_then(|swap| {
+                Ok(SwapInfo {
+                    status: swap.status(),
+                    lockup_address: swap.lockupAddress,
+                    expected_amount: swap
+                        .onchainAmount
```
michael1011: Why are you mapping the *actual* `onchainAmount` to the expected one?

boltzr/src/service/funding_address_signer.rs
```diff
@@ -142,6 +143,16 @@ impl FundingAddressSigner {
...
+        let funding_address_amount = funding_address.lockup_amount.ok_or_else(|| {
+            anyhow!(FundingAddressEligibilityError(
+                "funding address lockup amount missing".to_string(),
+            ))
+        })?;
+        if funding_address_amount != swap_info.expected_amount {
+            return Err(anyhow!(FundingAddressEligibilityError(
+                "funding address amount does not match swaps expected amount".to_string(),
+            )));
+        }
```
michael1011: Let's also validate that the chain we expect the swap to receive on is the chain of the funding address with a nice error

PR #1366, opened 2026-04-22
fix: harden swap validation and expiry handling

maybeast:
## Summary by CodeRabbit

* **New Features**
  * Enforced pagination limit (1–1000) for rescue requests.
  * Zero-confirmation handling added to swap renegotiation (acceptZeroConf propagated).

* **Bug Fixes**
  * Added "invoice.paid" swap status for improved tracking.

* **Documentation**
  * API/OpenAPI docs updated to reflect pagination limit (maximum 1000).

* **Refactor**
  * Consolidated and optimized swap filtering logic for invoice expiration checks.

### DIFF
--- a/boltzr/src/service/pubkey_iterator.rs
+++ b/boltzr/src/service/pubkey_iterator.rs
@@ -12,6 +12,7 @@ use std::{
 };
 
 pub const MAX_GAP_LIMIT: u32 = 150;
+pub const MAX_PAGINATION_LIMIT: u32 = 1000;
 
 const DEFAULT_GAP_LIMIT: u32 = 50;
 const DEFAULT_DERIVATION_PATH: &str = "m/44/0/0/0";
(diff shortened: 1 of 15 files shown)


boltzr/src/service/pubkey_iterator.rs
```diff
@@ -12,6 +12,7 @@ use std::{
 };
 
 pub const MAX_GAP_LIMIT: u32 = 150;
+pub const MAX_PAGINATION_LIMIT: u32 = 1000;
```
michael1011, line 15: https://github.com/BoltzExchange/boltz-backend/pull/1366/changes#r3125209355
Valid concern. Let's use checked math here

lib/service/Renegotiator.ts
```diff
@@ -69,6 +69,11 @@ class Renegotiator {
           throw Errors.INVALID_QUOTE();
         }
 
+        const acceptZeroConf = this.rateProvider.acceptZeroConf(
+          receivingCurrency.symbol,
+          swap.receivingData.amount!,
+        );
```
michael1011: Let's only do this check when the current `acceptZeroConf` value is true. Just to be on the safe side when renegotiating existing chain swap with amounts

michael1011 (changes requested):

michael1011 (approved):

PR #1371, opened 2026-04-23
feat: store found claim transactions ids

### LINKED ISSUE #1368: EVM: add user claim tx to `swapinfo`
Currently only available in logs.

michael1011:
Closes https://github.com/BoltzExchange/boltz-backend/issues/1368

Important limitation is that we do not scan for cooperative claim transactions on UTXO based chains because we already get the preimage from the cooperative signing flow. So, those will not be saved in the database.

## Summary by CodeRabbit

* **New Features**
  * Track and persist claim transaction details (transaction IDs and chain symbols) when claims occur.
  * Include claimTransactionId and transactionHash in swap notification payloads and claim events.
  * Add validation to ensure claim transactions reference existing swaps.

* **Tests**
  * Added extensive tests covering claim transaction persistence, retrieval, notification integration, and resilience when persistence fails.

### DIFF
--- a/lib/swap/EthereumNursery.ts
+++ b/lib/swap/EthereumNursery.ts
@@ -476,17 +476,17 @@ class EthereumNursery extends TypedEventEmitter<{
             `Found claim in ${this.ethereumManager.networkDetails.name} EtherSwap contract for ${swapTypeToPrettyString(swap.type)} Swap ${swap.id}: ${transactionHash}`,
           );
 
-          await ClaimTransactionRepository.addTransaction({
-            swapId: swap.id,
-            symbol: this.ethereumManager.networkDetails.symbol,
-            id: transactionHash,
-          });
-
           this.emit('claim', {
             swap,
             preimage,
             isEtherSwap: true,
             transactionHash,
+          });
+
+          await ClaimTransactionRepository.persistTransaction(this.logger, {
+            swapId: swap.id,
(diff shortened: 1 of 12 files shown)


lib/swap/EthereumNursery.ts
```diff
@@ -474,7 +476,18 @@ class EthereumNursery extends TypedEventEmitter<{
...
+            symbol: this.ethereumManager.networkDetails.symbol,
+            id: transactionHash,
+          });
+
+          this.emit('claim', {
+            swap,
+            preimage,
+            isEtherSwap: true,
+            transactionHash,
+          });
```
kiocos: I think `await ClaimTransactionRepository.addTransaction` should come after the `emit` and should be wrapped in try/catch, so we don't stop the claim handling in case there's an error writing to db.

kiocos (commented):
Don't we also need to write the claim tx id to db on `DeferredClaimer`? `broadcastClaim` currently doesn't do it

michael1011:
> Don't we also need to write the claim tx id to db on DeferredClaimer? broadcastClaim currently doesn't do it

This is about the claim transactions of the client. The cooperative ones we sign in the deferred claimer are intentionally out of scope because we can sign multiple claim transactions of theirs and don't scan the chain for which one actually got included

maybeast (approved):

PR #1372, opened 2026-04-23
fix: tolerate positive slippage

### LINKED ISSUE #1370: EVM: Swaps funded by external dex hop can exceed max
Due to positive slippage, swaps created near the max limit that are funded by an external DEX hop, e.g. USDT -> BTC can exceed our backend max amount. Possible solution: apply a `max(X sats, 2%)` buffer similar to what we are currently using as overpayment protection. Situation is different though, user is not actually overpaying, just being lucky. So we should be permissive.

michael1011:
Closes https://github.com/BoltzExchange/boltz-backend/issues/1370

Allow commitment swaps and chain swap renegotiation to accept positive slippage up to the configured overpayment tolerance.

## Summary by CodeRabbit

* **Documentation**
  * Added optional maxOverpaymentPercentage to commitment submission docs and clarified signature/amount constraints
  * Updated renegotiation guidance to reflect allowed overpayment within tolerance

* **New Features**
  * Introduced configurable positive-slippage tolerance for accepting overpayments
  * Locked amounts may now exceed previous maximal limits when within the configured tolerance

### DIFF
--- a/lib/service/Renegotiator.ts
+++ b/lib/service/Renegotiator.ts
@@ -1,3 +1,4 @@
+import type { OverPaymentConfig } from '../Config';
 import { parseTransaction } from '../Core';
 import type Logger from '../Logger';
 import { isTxConfirmed } from '../Utils';
@@ -20,6 +21,7 @@ import FeeProvider from '../rates/FeeProvider';
 import type RateProvider from '../rates/RateProvider';
 import { TransactionStatus } from '../sidecar/Sidecar';
 import ErrorsSwap from '../swap/Errors';
+import { getAllowedPositiveSlippageFromConfig } from '../swap/OverpaymentProtector';
 import type SwapNursery from '../swap/SwapNursery';
 import type { Currency } from '../wallet/WalletManager';
 import type WalletManager from '../wallet/WalletManager';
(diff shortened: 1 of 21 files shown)


michael1011:
@coderabbitai review

lib/service/Renegotiator.ts
```diff
@@ -268,7 +271,14 @@ class Renegotiator {
...
+      swap.receivingData.amount! >
+      pair.limits.maximal +
+        getAllowedPositiveSlippageFromConfig(
+          pair.limits.maximal,
+          this.overPaymentConfig,
+        )
+    ) {
       throw Errors.EXCEED_MAXIMAL_AMOUNT(
         swap.receivingData.amount!,
         pair.limits.maximal,
```
kiocos, line 282: should be maximal + tolerance

kiocos (changes requested):

lib/service/Renegotiator.ts
michael1011: This is intentional. We don't want to advertize our soft limits

lib/swap/OverpaymentProtector.ts
```diff
@@ -2,27 +2,60 @@ import type { OverPaymentConfig } from '../Config';
...
+export const getAllowedPositiveSlippageFromConfig = (
+  amount: number,
+  config?: OverPaymentConfig,
+  maxPercentageOverride?: number,
+): number =>
+  getAllowedPositiveSlippage(
+    amount,
+    resolvePositiveSlippageTolerance(config, maxPercentageOverride),
+  );
+
```
jackstar12: We might aswell put all these into the `OverpaymentProtector` and pass that around - its only a wrapper around the config + logger anyways.

jackstar12 (approved):

maybeast (approved):

PR #1389, opened 2026-05-05
feat: 0 amount commitments for chain swaps

jackstar12:
## Release Notes

* **Bug Fixes**
  * Improved amount validation so certain swaps (including amountless chain cases) only validate amounts when appropriate, reducing false overpayment/insufficient errors.
* **Tests**
  * Added unit and integration tests covering amountless chain swaps and zero-amount chain lockup behaviors to ensure correct emissions and persistence.

### DIFF
--- a/lib/swap/EthereumNursery.ts
+++ b/lib/swap/EthereumNursery.ts
@@ -229,7 +229,7 @@ class EthereumNursery extends TypedEventEmitter<{
     }
 
     const expectedAmount = this.getSwapExpectedReceivingAmount(swap);
-    if (expectedAmount) {
+    if (this.shouldValidateAmount(swap, expectedAmount)) {
       const actualAmountSat = Number(etherSwapValues.amount / etherDecimals);
 
       if (
@@ -384,7 +384,7 @@ class EthereumNursery extends TypedEventEmitter<{
     }
 
     const expectedAmount = this.getSwapExpectedReceivingAmount(swap);
-    if (expectedAmount) {
+    if (this.shouldValidateAmount(swap, expectedAmount)) {
       const actualAmount = erc20Wallet.normalizeTokenAmount(
         erc20SwapValues.amount,
       );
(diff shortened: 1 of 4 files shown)


lib/swap/EthereumNursery.ts
```diff
@@ -720,6 +720,15 @@ class EthereumNursery extends TypedEventEmitter<{
       ? (swap as Swap).expectedAmount
       : (swap as ChainSwapInfo).receivingData.expectedAmount;
 
+  private shouldValidateAmount = (
+    swap: Swap | ChainSwapInfo,
+    expectedAmount: number | null | undefined,
+  ): expectedAmount is number => {
+    // Submarine swaps can be locked up before the invoice is set; amount
+    // validation happens after the invoice populates the expected amount.
+    return Boolean(expectedAmount) || swap.type === SwapType.Chain;
```
michael1011: `Boolean(expectedAmount)` being implicit like this is what got us in this mess in the first place. Let's write the conditions here *explicitly*. `expectedAmount is number` is not checked either

test/integration/wallet/ethereum/contracts/Commitments.spec.ts
```diff
@@ -950,6 +1002,69 @@ describe('Commitments', () => {
       );
     });
 
+    test('should create ERC20 commitment for amountless chain swap', async () => {
```
michael1011, line 1010: There is not test for Ether

lib/wallet/ethereum/contracts/Commitments.ts
```diff
@@ -313,8 +315,16 @@ class Commitments {
...
     expectedAmount: number,
     actualAmount: number,
+    swapType: SwapType,
     maxOverpaymentPercentage?: number,
   ) => {
+    if (swapType === SwapType.Chain && expectedAmount === 0) {
+      if (actualAmount <= 0) {
+        throw new Error('commitment amount has to be greater than 0');
+      }
+      return;
```
michael1011, line 322: Why does this check exist? The contracts shouldn't allow you to lock 0

michael1011 (changes requested):

michael1011 (approved):

PR #1429, opened 2026-06-04
feat: hold in invoice payment hook

maybeast:
## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the intermediate "Hold" action from swap processing so lockup events are emitted immediately instead of being deferred.
  * Invoice payments now respect explicit Hold vs Continue results, so payments can be paused when a hold is returned.

* **Tests**
  * Updated and expanded unit tests to reflect Hold removal and to cover transaction hooks, invoice payment hooks, node selection, and swap retry behavior.

### DIFF
--- a/lib/swap/NodeSwitch.ts
+++ b/lib/swap/NodeSwitch.ts
@@ -17,6 +17,8 @@ import {
 import Errors from './Errors';
 import InvoicePaymentHook, {
   InvoicePaymentHookAction,
+  type InvoicePaymentHookContinue,
+  type InvoicePaymentHookHold,
 } from './hooks/InvoicePaymentHook';
 
 type NodeAmountThreshold = {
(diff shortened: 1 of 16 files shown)


lib/swap/hooks/InvoicePaymentHook.ts
```diff
@@ -77,13 +110,33 @@ class InvoicePaymentHook extends Hook<
...
     }
 
-    return { nodeId };
+    return { action, nodeId };
+  };
+
+  private parseHookAction = (
+    res: boltzrpc.InvoicePaymentHookResponse,
+  ): InvoicePaymentHookAction | undefined => {
+    switch (res.action ?? boltzrpc.InvoicePaymentHookAction.CONTINUE) {
```
michael1011: We shouldn't define defaults in multiple places. Resuse `defaultHookResult.action` here

lib/swap/NodeSwitch.ts
```diff
@@ -51,6 +53,25 @@ type ReverseSwapNodeResolution =
       reason: string;
     };
 
+enum InvoicePaymentDecision {
```
michael1011: Duplicate of `InvoicePaymentHookAction`

lib/swap/NodeSwitch.ts
```diff
@@ -51,6 +53,25 @@ type ReverseSwapNodeResolution =
...
+
+type InvoicePaymentHookContinue = {
+  action: InvoicePaymentDecision.Continue;
+  client?: LightningClient;
+  timePreference?: number;
+};
+
+type InvoicePaymentHookResult =
+  | InvoicePaymentHookHold
+  | InvoicePaymentHookContinue;
```
michael1011: These things shouldn't be defined in the consumer of the hook but by the hook itself

michael1011 (changes requested):

michael1011 (approved):

PR #1430, opened 2026-06-04
Payjoin integration

pseudozach:
This PR revives PR #892 against latest rust-payjoin, using the current receive v2 state machine and DB-backed event persistence. It adds a Rust-side Payjoin v2 receiver flow to submarine swaps and wires it into BTC -> LN swap creation. The goal is to preserve the existing public API shape while making the existing `bip21` field a Payjoin-capable BIP21 URI when Payjoin is available.

The implementation is based on the current `rust-payjoin` receive v2 typestate API. It persists rust-payjoin session events and replays those events to recover receiver state.

At a high level, the flow is:

1. BTC -> LN submarine swap creation builds the normal lockup address and expected amount.
2. `Service.setSwapInvoice` builds a normal BIP21 first.
3. For BTC -> BTC-lightning submarine swaps only, it asks the Rust sidecar for a Payjoin URI using the same address, amount, label, and swap id.
4. If Payjoin URI generation succeeds, the response `bip21` is replaced with the Payjoin URI.
5. If Payjoin generation fails, Boltz falls back to the normal BIP21.
6. The Rust sidecar creates a Payjoin receiver session, persists all rust-payjoin events, processes the sender proposal, posts the Payjoin proposal, monitors payment, and closes the session.
7. TypeScript swap validation uses the linked Payjoin session to allow the intentional oversized consolidation lockup only when it is proven to be the finalized Payjoin transaction for that swap.

### Existing `bip21` field is reused
For eligible swaps, `bip21` becomes:
```text
bitcoin:<same-address>?amount=<same-amount>&label=<same-label>&pj=...
```
For non-Payjoin swaps, or if Payjoin generation fails, `bip21` remains the normal non-Payjoin URI.

### Payjoin generation lives in `Service.setSwapInvoice`
This keeps normal behavior unchanged if Payjoin is unavailable.

### Receiver state is event-sourced
The Rust receiver persists rust-payjoin `SessionEvent`s into:
- `payjoinReceiverSessions`
- `payjoinReceiverSessionEvents`

### Swap linkage
The rust-payjoin event log only stores Payjoin protocol events. Boltz-specific metadata, such as the swap id, lives on `payjoinReceiverSessions`.

This keeps the protocol event stream replayable by rust-payjoin without mixing in Boltz-only fields, while still allowing Boltz to enforce:

```text
one swap -> at most one Payjoin receiver session
```

### Seen-input tracking is persisted globally
This follows the rust-payjoin `payjoin-cli` pattern and is required for automated receivers. It prevents [probing attempts](https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki#user-content-span_idprobingattackspanOn_the_receiver_side_UTXO_probing_attack) where a sender reuses the same original input to repeatedly solicit receiver contributions and learn about Boltz wallet UTXOs. 

### OHTTP relay selection follows payjoin-cli
The sidecar uses a small runtime `RelayManager` modeled after `payjoin-cli`.
It:
- randomly selects from configured relays
- remembers the selected relay
- excludes failed relays
- reuses the selected relay for receiver poll/post requests
- returns immediately on `UnexpectedStatusCode`, matching the reference behavior

### Oversized Payjoin lockups are accepted only when proven
Payjoin submarine swaps intentionally create a large lockup output when Boltz contributes one of its own UTXOs and consolidates it into the swap output. That means normal overpayment validation would reject the lockup.

The oversized lockup is accepted only if:
- the swap has a linked Payjoin receiver session
- the session has reached `Closed`
- the closed outcome is `Success`
- the observed lockup txid matches the txid reconstructed from the persisted `FinalizedProposal`
- the actual lockup amount is greater than the expected amount

If the session exists but is not closed yet, the lockup is deferred and retried. This handles the mempool race where UtxoNursery sees the transaction before rust-payjoin monitoring has persisted `Closed`.

-> Feedback is needed here from boltz team, if this is preferred approach or would rather do this differently.

### Signing
`finalize_proposal` signs through the existing BTC chain client by calling Bitcoin Core wallet PSBT processing. The signing path is kept isolated in `sign_payjoin_psbt`.

### Monitoring
`Receiver<Monitor>::check_payment` uses the existing BTC chain client to look up the Payjoin transaction. It checks both general raw transaction lookup and wallet transaction lookup because some relevant transactions may be wallet-known before they are generally queryable.

-> feedback welcome here, as there's some additional delayed checks are introduced

### Relay/directory configuration
Defaults are currently in the Rust module with env var overrides:
- `BOLTZ_PAYJOIN_DIRECTORY`
- `BOLTZ_PAYJOIN_OHTTP_RELAYS`

### UTXO contribution policy
The current input selection picks the smallest spendable wallet UTXO that rust-payjoin accepts. This is intentionally minimal and isolated.

### Seen-input retention
Seen inputs are stored permanently for now. This is the conservative anti-probing behavior.

--- 

here's a demo video
https://github.com/user-attachments/assets/9eb9ac70-6f00-4d76-85b1-35b780b447c5

note: I've utilized help of codex-5.5 in writing the code and the summary of this PR.


DanGould:
excellent video. I wonder if the swap is normally only detected after confirmation or if the normal detection code is missing the transaction because it's now multi input/multi-output shaped. We should confirm before merge

other small things:

1.0.0-rc2 is older than 0.25. Better to switch to 0.25 or just work off of master until we cut another 1.0.0-rc.

Seems like minus lock files this is only around 1500+ lines. Love to see it.

maybeast:
Is this ready for review?

DanGould:
not yet imoh we gotta bring it up to the same standard as the reference & other impls

lib/service/Service.ts
```diff
@@ -1561,16 +1562,19 @@ class Service {
       );
 
     await this.createExtraFees(swap.id, fees.extraFee, extraFees);
+    const label = getSwapMemo(lightningCurrency, SwapType.Submarine);
```
DanGould, line 1567: getPayjoinUri makes more sense to me here because it's a network call and will let `encodeBip21` remain a pure function.

boltzr/src/db/models/payjoin.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Queryable, Selectable, AsChangeset, PartialEq, Clone, Debug)]
+#[diesel(table_name = crate::db::schema::payjoinReceiverSessions)]
+#[allow(non_snake_case)]
+pub struct PayjoinReceiverSession {
+    pub id: i64,
+    pub swapId: Option<String>,
+    pub address: String,
+    pub amountSats: Option<i64>,
+    pub label: Option<String>,
+    pub payjoinTxId: Option<String>,
```
DanGould: This shouldn't be necessary because you can get it from the [`Receiver<Monitor>::check_payment` ](https://docs.rs/payjoin/1.0.0-rc.3/payjoin/receive/v2/struct.Receiver.html#impl-Receiver%3CMonitor%3E)function. That takes a closure that'll give you the strongly typed TxId so that you don't need to save a String and you don't need to save it in the db at all

boltzr/src/payjoin/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+                if transaction.is_some() {
+                    let mut detected = detected_payjoin_txid.lock().map_err(|_| {
+                        implementation_error("failed to lock detected payjoin txid")
+                    })?;
+                    *detected = Some(txid);
+                }
+
+                Ok(transaction)
+            })
+            .save(persister)
```
DanGould, line 648: `detected_payjoin_txid` isn't doing anything here. You can get rid of it because the txid is given to you there in the closure, in |txid|.

boltzr/src/payjoin/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+        .filter(|relays| !relays.is_empty())
+        .unwrap_or_else(|| OHTTP_RELAYS.iter().map(|relay| relay.to_string()).collect())
+}
+
+fn payjoin_ohttp_relay() -> String {
+    payjoin_ohttp_relays()
+        .into_iter()
+        .next()
+        .unwrap_or_else(|| OHTTP_RELAY.to_string())
+}
```
DanGould, line 787: We definitely want the best practice relay selection (see payjoin-cli for reference, spec coming soon) before this merges

boltzr/src/payjoin/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+    persister: &ReceiverPersister,
+    btc_chain: Arc<dyn Client + Send + Sync>,
+    receiver_address: Address,
+    network: payjoin::bitcoin::Network,
+) -> Result<()> {
+    let receiver = receiver
+        .check_no_inputs_seen_before(&mut |_outpoint| {
+            // TODO: replace with DB-backed seen-input tracking before enabling full proposal flow.
+            Ok(false)
+        })
```
DanGould: This needs to be implemented to prevent probing attacks because Boltz is an automated receiver.

boltzr/src/payjoin/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+    let mut errors = Vec::new();
+
+    for relay in payjoin_ohttp_relays() {
+        match payjoin::io::fetch_ohttp_keys(relay.as_str(), directory).await {
+            Ok(keys) => return Ok(keys),
+            Err(error) => {
+                warn!(
+                    relay = relay.as_str(),
+                    error = %error,
+                    "Failed to fetch Payjoin OHTTP keys from relay"
```
pseudozach: keys are fetched from directory

DanGould (changes requested):

boltzr/src/payjoin/mod.rs
pseudozach: done, removed.

boltzr/src/payjoin/mod.rs
pseudozach: implemented, would love feedback as part of review.

PR #1434, opened 2026-06-08
feat: rescue EVM swaps

michael1011:
## Summary by CodeRabbit

* **New Features**
  * Added restoring swaps via an EVM claim address using a timestamp + signature ownership proof.
  * Extended swap rescue/restore responses with polymorphic claim details supporting both UTXO and EVM claims, including EVM transaction metadata.

* **Bug Fixes**
  * Improved restore result ordering so swaps missing legacy key-index information are listed last.

* **Documentation**
  * Updated the `/swap/restore` API contract and schemas: request now uses `RestoreRequest`, transactions identify via `id` + `vout`, claim/refund details require `timeoutBlockHeight` (and refunds require `serverPublicKey`), and EVM claim schemas were added.


kiocos (approved):

boltzr/src/service/rescue.rs
```diff
@@ -365,6 +412,8 @@ pub struct RestorableSwap {
     pub claim_details: Option<ClaimDetails>,
     #[serde(rename = "refundDetails", skip_serializing_if = "Option::is_none")]
     pub refund_details: Option<SwapDetailsBase>,
+    #[serde(rename = "evmClaimDetails", skip_serializing_if = "Option::is_none")]
+    pub evm_claim_details: Option<EvmDetails>,
```
jackstar12: Making `claim_details` an enum with the different implementation types and have it as a `oneOf` in the openapi spec would make for a nicer api imho

boltzr/src/service/rescue.rs
```diff
@@ -1659,6 +1786,7 @@ mod test {
                     preimage_hash: reverse_swap.preimageHash.clone(),
                 }),
                 refund_details: None,
+                evm_claim_details: None,
```
jackstar12: Let's fill in all of these values with the defaults (all the cases below too)

boltzr/src/utils/serde.rs
```diff
@@ -79,6 +80,81 @@ impl<'de> Deserialize<'de> for ElementsAddressDeserialize {
...
+                formatter.write_str("an array of EVM addresses")
+            }
+
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::SeqAccess<'de>,
+            {
+                let mut vec = Vec::new();
+                while let Some(value) = seq.next_element::<String>()? {
+                    if vec.len() >= MAX_GAP_LIMIT as usize {
```
jackstar12: Does it make sense to use the gap limit const here? Maybe we should consider this as a seperate limit?

maybeast (changes requested):
no lookup by evm address

lib/api/v2/routers/SwapRouter.ts
```diff
@@ -2274,6 +2344,12 @@ class SwapRouter extends RouterBase {
      *               type: array
      *               items:
      *                 $ref: '#/components/schemas/RestorableSwap'
+     *       '400':
```
maybeast, line 2347: The Rust restore endpoint returns 422 for invalid EVM restore inputs such as signature mismatch, stale timestamp, and invalid signature/address. The tests assert 422 in `boltzr/src/api/rescue.rs`

boltzr/src/api/rescue.rs
```diff
@@ -32,13 +34,51 @@ pub enum RescueParams {
...
+    format!("Boltz swap restore\naddress: {address}\ntimestamp: {timestamp}")
+}
+
+fn verify_restore_signature(
+    address: &boltz_evm::Address,
+    timestamp: u64,
+    signature: &str,
+) -> Result<(), AxumError> {
+    let now = chrono::Utc::now().timestamp();
+    if (now - timestamp as i64).abs() > RESTORE_SIGNATURE_VALIDITY_SECS {
```
maybeast: This freshness check casts a user-controlled `u64` timestamp to `i64` before subtracting. Very large values can wrap, and the subtraction/abs path can behave unexpectedly around integer limits. It does not bypass address ownership, but it weakens the replay guard on a privacy-sensitive restore proof.

 Can we make this explicit, e.g. reject timestamps above `i64::MAX` and use `abs_diff`/checked arithmetic before signature recovery?

maybeast (approved):

kiocos (approved):

michael1011:
@coderabbitai review

PR #1445, opened 2026-06-22
feat: alert on claim fails

michael1011:
## Release Notes

* **New Features**
  * Added immediate “claim failure” notifications for all supported swap types, including swap identifiers and concise error details (errors are truncated to 200 characters when needed).
  * Standardized claim-failure events so notification delivery is triggered reliably whenever claim operations fail.

* **Tests**
  * Expanded unit test coverage to verify notification message formatting, error truncation behavior, and that success events are not emitted on claim failures.

### DIFF
--- a/lib/notifications/NotificationProvider.ts
+++ b/lib/notifications/NotificationProvider.ts
@@ -302,6 +302,22 @@ class NotificationProvider {
       );
     });
 
+    this.service.eventHandler.on(
+      'claim.failure',
+      async ({ swap, symbol, error }) => {
+        const { base, quote } = splitPairId(swap.pair);
+        const truncatedError =
+          error.length > 200 ? error.substring(0, 200) + '...' : error;
+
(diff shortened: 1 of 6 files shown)


lib/notifications/NotificationProvider.ts
```diff
@@ -302,6 +302,22 @@ class NotificationProvider {
...
     });
 
+    this.service.eventHandler.on(
+      'claim.failure',
+      async ({ swap, symbol, error }) => {
+        const { base, quote } = splitPairId(swap.pair);
+        const truncatedError =
+          error.length > 200 ? error.substring(0, 200) + '...' : error;
+
+        const message =
```
jackstar12, line 312: Symbol is being duplicated in the error message here, already included in the formatted error message. Doesn't hurt but could remove in either place.

jackstar12 (approved):

PR #1455, opened 2026-07-02
feat: add PATCH endpoint for swap metadata

kiocos:
## Summary by CodeRabbit

* **New Features**
  * Added `PATCH /swap/{id}/metadata` to set or replace swap metadata.
* **Bug Fixes**
  * Improved metadata update validation (request shape and HEX/size rules) and now returns 404 when the swap can’t be found.
* **Documentation**
  * Updated Swagger to document the new endpoint and introduced reusable metadata request/HEX schemas.
* **Tests**
  * Expanded unit and integration coverage for metadata updates, including overwrite behavior and validation/error handling.

### DIFF
--- a/lib/api/v2/routers/SwapRouter.ts
+++ b/lib/api/v2/routers/SwapRouter.ts
@@ -364,17 +364,17 @@ class SwapRouter extends RouterBase {
 
     /**
      * @openapi
-     * /swap/submarine/{id}/metadata:
+     * /swap/{id}/metadata:
      *   patch:
-     *     tags: [Submarine Swap]
-     *     description: Set or replace the metadata stored alongside a Submarine Swap
+     *     tags: [Swap]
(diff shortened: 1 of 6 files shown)


lib/api/v2/routers/SwapRouter.ts
```diff
@@ -343,6 +347,65 @@ class SwapRouter extends RouterBase {
...
+     *         description: No swap with the provided ID exists
+     *         content:
+     *           application/json:
+     *             schema:
+     *               $ref: '#/components/schemas/ErrorResponse'
+     */
+    router.patch(
+      '/submarine/:id/metadata',
+      this.handleError(this.patchSubmarineMetadata),
+    );
```
michael1011: Let's not do a new endpoint for every swap type but just `/:id/metadata`

michael1011 (changes requested):

PR #1458, opened 2026-07-05
feat: signed swap-response attestation for reverse swaps

kaloudis:
## Motivation

Any Boltz integration where the entity that terminates the TLS connection to boltz.exchange is a separate process from the one holding the customer's spending key has an exfil vector: the coordinator can substitute the client's `claimPublicKey` (reverse) or `refundPublicKey` (submarine) when relaying the create-swap request, and neither the Bolt11 invoice signature nor the TLS certificate binds that substitution. The signer sees an invoice it recognizes and approves the LN payment; Boltz builds the on-chain HTLC against the attacker's key; the coordinator sweeps.

This class of split-signer architecture is becoming common — self-custodial wallet vendors, routing-node control planes, custody products with in-enclave signers. We're building one for the ZEUS routing-node product (signer in a Nitro enclave, coordinator on a customer's VPS or self-hosted server) and we can't safely allowlist Boltz payees for our approver's short-circuit without this.

TLS doesn't help — TLS proves the response to whoever terminates the connection, which is the coordinator, not the signer behind it. The Bolt11 invoice signature already binds `(payment_hash, amount)` but explicitly not `claim_pubkey` / `refund_pubkey` / `lockup_address`. A payload signature over those fields under the same LN node key is the missing bind.

## What this PR does

Reverse-swap creation responses (`POST /v2/swap/reverse`) grow a new optional field:

```json
{
  "id": "...",
  "invoice": "lnbc...",
  "lockupAddress": "bc1p...",
  "swapTree": { ... },
  ...
  "attestation": {
    "version": 1,
    "signature": "zbase32-encoded-compact-ecdsa-with-recovery"
  }
}
```

`signature` is the output of `signMessage` under the LN node key associated with the swap (LND's `lnrpc.SignMessage` / CLN's `signmessage`). The signed input is the lowercase hex encoding of a SHA256 commitment to `(tag || swap_id || payment_hash || client_pubkey || lockup_address || timeout || onchain_amount)`. The commitment layout, verifier pseudocode, and version-negotiation semantics are documented in `docs/swap-response-attestation.md`.

Verifiers already have the signing key — it's the same one that signs the Bolt11 invoice returned in the response. No new pinning ceremony, no additional trust anchor.

## Scope + non-scope

- **In scope**: reverse-swap attestation for UTXO-based swaps where the client supplied a `claimPublicKey`. Small additive change per swap; no schema break for existing clients.
- **Out of scope for this PR**: submarine swaps. The LN node that will pay the invoice isn't picked until the client submits their invoice via `setInvoice`, which is a natural follow-up hook point. Happy to open a second PR against `Service.setInvoice` if this shape lands.
- **Out of scope**: MuSig2 cooperative-claim protection. The attestation binds the script-path spend key, which is the only path a compromised coordinator can drive on its own.

## Design choices worth flagging

- **LN node key vs a dedicated swap-signing key.** Chose the node key because clients already fetch + verify it for the invoice, and it kept the key-custody policy uniform on Boltz's side (existing HSM, rotation cadence). A dedicated key would add a distribution + rotation story for zero additional security.
- **zbase32 wire format.** Both `lnrpc.SignMessage` and `cln.signmessage` natively emit zbase32; forwarding it verbatim keeps the LN-Signed-Message convention. Verifiers can decode with any standard zbase32 library.
- **hex-encoded commitment on the wire to `signMessage`.** CLN's `signmessage` proto types the message as a string; hex is UTF-8-safe and matches what LND signs when handed the same string. Documented as an interface constraint on `LightningClient.signMessage`.

## Test plan

- [x] `npx jest test/unit` — all 1476 unit tests pass (13 new)
- [x] `npx tsc --noEmit` — clean
- [ ] Integration test against a live LND + CLN instance — I didn't wire this in the PR since your CI framework is more standardized; happy to add if you'd prefer

### DIFF
--- a/docs/swap-response-attestation.md
+++ b/docs/swap-response-attestation.md
@@ -1,0 +1,142 @@
+# Swap-response attestation
+
+Reverse swap responses include a Boltz-node-signed commitment to the on-chain
+HTLC parameters. This lets a client whose signer sits behind a separate
+coordinator process — a routing-node control plane, a wallet backend, a
+signer-in-enclave setup — cryptographically verify that the swap's
+`claimPublicKey`, `lockupAddress`, `timeoutBlockHeight` and `onchainAmount`
+weren't tampered with by the coordinator on the way to the signer.
+
+TLS is not sufficient here: TLS proves things to the endpoint that terminates
+the connection. If that endpoint is a compromised coordinator forwarding to a
+signer behind it, the signer sees only what the coordinator chooses to relay. A
+signature under the LN node key survives the coordinator unchanged because it's
+anchored to the same key that signs the swap's Bolt11 invoice.
+
+## Response shape
+
+Present on the reverse-swap creation response
+([`POST /v2/swap/reverse`](./api-v2.md)) whenever the client supplied a
+`claimPublicKey` for a UTXO-based reverse swap:
+
+```json
+{
+  "id": "...",
+  "invoice": "lnbc...",
+  "lockupAddress": "bc1p...",
+  "swapTree": { ... },
+  "refundPublicKey": "02...",
+  "timeoutBlockHeight": 820000,
+  "onchainAmount": 250000,
+  "attestation": {
+    "version": 1,
+    "signature": "zbase32-encoded-compact-ecdsa-with-recovery"
+  }
+}
+```
+
+## Commitment layout
+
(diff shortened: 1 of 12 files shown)


kaloudis:
Closing this out in favor of a client-side design where no Boltz backend changes are needed.

PR #1462, opened 2026-07-11
fix: confirm vHTLC refunds via RefundWatcher

michael1011:
## Summary by CodeRabbit

* **New Features**
  * Added an ARK CLI **Decode** command that accepts a base64 PSBT or a file path and outputs decoded transaction details as pretty JSON.

* **Bug Fixes**
  * Prevented duplicate refund confirmations by only transitioning refunds from **Pending** to **Confirmed** once, and avoiding “confirmed” events when no update occurs.
  * Updated ARK vtxo refund flow to emit **pending** first, then rely on monitoring to emit the **confirmed** event.
  * Refund monitoring now continues processing other pending items even if individual checks fail.

* **Tests**
  * Expanded unit and integration tests to cover ARK decoding and the pending→confirmed refund status transition behavior.

### DIFF
--- a/lib/swap/SwapNursery.ts
+++ b/lib/swap/SwapNursery.ts
@@ -53,7 +53,6 @@ import type {
   ERC20SwapValues,
   EtherSwapValues,
 } from '../consts/Types';
-import { RefundStatus } from '../db/models/RefundTransaction';
 import type ReverseSwap from '../db/models/ReverseSwap';
 import type SendApprovalHold from '../db/models/SendApprovalHold';
 import type Swap from '../db/models/Swap';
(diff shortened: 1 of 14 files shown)


lib/swap/SwapNursery.ts
```diff
@@ -2536,24 +2535,29 @@ class SwapNursery extends TypedEventEmitter<SwapNurseryEvents> {
...
-      swap: await WrappedSwapRepository.setTransactionRefunded(
-        swap,
-        0,
-        Errors.REFUNDED_COINS(txId).message,
-      ),
+      swap: refundedSwap,
     });
+
+    // Ark transactions are final instantly, so the watcher can confirm the
+    // refund in isolation instead of waiting for or blocking on a global sweep
```
jackstar12:
```suggestion
    // Ark transactions are final instantly, so confirm via the watcher directly
    // instead of waiting on the global pending sweep. The refund must still flow
    // through RefundWatcher so downstream handlers (e.g. invoice cancellation on
    // refund.confirmed) run on the same path as other chains.
    
```

jackstar12 (approved):

PR #1464, opened 2026-07-14
feat: add generic failure hook

captainpushalot:
## Summary

- add one authenticated bidirectional `FailureHook` stream with a typed `oneof` payload envelope
- make claim failures the first payload type, carrying immediate or batch type, symbol, optional swap ID, and batch size
- correlate each delivered event and acknowledgement with a UUID so acknowledgements can arrive out of order
- keep claim execution independent of subscriber availability, acknowledgement timeouts, and disconnects

Future failure categories can extend the envelope without adding another stream or RPC. With no subscriber connected the hook is a no-op, and the claim path never awaits acknowledgement. The contract is best-effort and does not provide durable replay.

## Testing

- protobuf generation and TypeScript/Rust compilation
- focused hook, service, gRPC, and deferred-claimer tests: 239 passing
- full TypeScript unit suite: 109 suites, 2,357 tests, 20 snapshots passing
- formatting, lint, Swagger generation/validation, `cargo fmt`, and `cargo clippy`

## Summary by CodeRabbit

* **New Features**
  * Added a gRPC FailureHook stream to receive claim-failure notifications.
  * Notifications now include structured failure details with `ClaimFailureType` (immediate vs batch), including asset symbol plus either swap ID or batch size.
* **Tests**
  * Added/updated unit and integration coverage for FailureHook streaming behavior, acknowledgements, and correct forwarding of immediate and batch claim failures (including batch-size assertions).

### DIFF
--- a/lib/grpc/GrpcServer.ts
+++ b/lib/grpc/GrpcServer.ts
@@ -60,7 +60,7 @@ class GrpcServer {
       invoiceCreationHook: grpcService.invoiceCreationHook,
       transactionHook: grpcService.transactionHook,
       sendApprovalHook: grpcService.sendApprovalHook,
-      claimFailureHook: grpcService.claimFailureHook,
+      failureHook: grpcService.failureHook,
       invoicePaymentHook: grpcService.invoicePaymentHook,
       getReferrals: grpcService.getReferrals,
       setReferral: grpcService.setReferral,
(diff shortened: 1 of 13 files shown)


michael1011 (commented):
On a higher level, I would prefer some generic failure hook over adding a new method for every specific kind of failure

captainpushalot:
@michael1011 Addressed in `b873bde9`: this is now a single `FailureHook` RPC with a typed `oneof` payload envelope. Claim failure is the first payload, so future failure categories extend the envelope instead of adding another stream or RPC.

The UUID event ID/acknowledgement contract remains for matching acknowledgements to in-flight events, including out-of-order responses. Claim execution still does not await the acknowledgement; no subscriber, disconnect, or timeout remains a no-op for the claim path.

PR #1468, opened 2026-07-17
feat: index submarine swaps on EVM address-based restore

kiocos:
Indexes submarine swaps on address-based /restore, so we can use the swap metadata to refund the locked TBTC to the original source.

## Summary by CodeRabbit

* **New Features**
  * Added optional on-chain refund address support for swaps and persisted it from submarine lockup events; refund-address-based restoration now returns typed refund details for both UTXO and EVM.
* **Documentation**
  * Updated swap restore OpenAPI/Swagger to describe EVM matching across claim and refund addresses with discriminated refund-details responses (UTXO vs EVM).
* **Bug Fixes**
  * Improved submarine swap recovery so refund-address restores yield correctly typed refund detail shapes.
* **Tests**
  * Added/updated repository, service, and API tests covering refund-address storage, overwrite behavior, and submarine EVM restoration.

### DIFF
--- a/lib/api/v2/routers/SwapRouter.ts
+++ b/lib/api/v2/routers/SwapRouter.ts
@@ -277,9 +277,6 @@ class SwapRouter extends RouterBase {
      *         refundPublicKey:
      *           type: string
      *           description: Public key with which the Submarine Swap can be refunded encoded as HEX
-     *         refundAddress:
-     *           type: string
-     *           description: EVM address with which the Submarine Swap can be refunded. Only allowed for swaps from EVM based chains
      *         pairHash:
      *           type: string
      *           description: Pair hash from the pair information for the client to check if their fee data is up-to-date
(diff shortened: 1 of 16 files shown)


lib/api/v2/routers/SwapRouter.ts
```diff
@@ -277,6 +277,9 @@ class SwapRouter extends RouterBase {
      *         refundPublicKey:
      *           type: string
      *           description: Public key with which the Submarine Swap can be refunded encoded as HEX
+     *         refundAddress:
+     *           type: string
+     *           description: EVM address with which the Submarine Swap can be refunded. Only allowed for swaps from EVM based chains
```
michael1011: Don't set this on create. Set this when we actually see the lockup tx

michael1011 (changes requested):

michael1011 (approved):

PR #1483, opened 2026-07-26
fix: swap update event races

michael1011:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lockup tracking by matching the exact on-chain output using transaction output/vout and Ethereum log index.
  * Prevented competing or unauthorized lockup updates from overwriting existing records, including idempotent retries.
  * Avoided confirmed→non-confirmed downgrades and tightened zero-confirmation rejection handling.
  * Refined Ethereum lockup verification and reduced incorrect/duplicate lockup failure emissions.
* **Reliability**
  * Serialized concurrent lockup validation per swap and improved event deduplication while preserving distinct lockup events.
  * Propagated log-index/output details through lockup, claim, and settlement flows.

### DIFF
--- a/lib/db/repositories/ChainSwapRepository.ts
+++ b/lib/db/repositories/ChainSwapRepository.ts
@@ -13,6 +13,12 @@ import {
 } from '../../consts/Enums';
 import type { IncorrectAmountDetails } from '../../consts/Types';
 import Database from '../Database';
+import type { LockupTargetStatus, LockupWriteResult } from '../LockupIdentity';
+import {
+  LockupWriteOutcome,
+  decideLockupWrite,
+  shouldWriteZeroConfRejection,
+} from '../LockupIdentity';
 import type { ChainSwapType } from '../models/ChainSwap';
 import ChainSwap from '../models/ChainSwap';
 import type { ChainSwapDataType } from '../models/ChainSwapData';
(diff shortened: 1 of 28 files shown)


lib/db/repositories/ChainSwapRepository.ts
```diff
@@ -360,6 +360,7 @@ class ChainSwapRepository {
         });
         const currentData = await ChainSwapData.findOne({
           transaction,
+          lock: transaction.LOCK.UPDATE,
```
jackstar12, line 363: its annoying we have to remember to set that cant we lock on the transaction level?

maybeast (approved):

lib/swap/ArkNursery.ts
```diff
@@ -38,6 +38,7 @@ class ArkNursery extends TypedEventEmitter<{
   'swap.lockup': {
     swap: Swap;
     lockupTransactionId: string;
+    lockupTransactionVout: number;
```
jackstar12, line 41: we should reuse the `LockupIdentity` primitives we've started building + swap type helpers for easy comparison and checks on the identity. could be a potential followup tho.

lib/swap/EthereumNursery.ts
```diff
@@ -180,6 +180,13 @@ class EthereumNursery extends TypedEventEmitter<{
       recordedStatus: swap.status as SwapUpdateEvent,
     });
 
+  private static shouldEmitLockupFailed = (
```
jackstar12, line 183: instead of dealing with the edge cases where we haven't persisted the failure reason why dont we start to atomically persist failure reasons with the lockup failed status? the db handlers could accept a target type that also accepts a failure reason for the lockup failed event

jackstar12 (approved):

